PR #101, opened 2018-11-28
RFC: Make this into a gem for improved rubocop usage?

shawnfrench:
Making `ruby-style-guide` into a gem would improve our development workflow with rubocop at Shopify.

### How?
Once the dependency is added:
`gem 'ruby-style-guide', '~> 0.0.1', source: PACKAGE_CLOUD`

the Shopify project can use the `inherit_gem` config directive (documented [here](https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#inheriting-configuration-from-a-dependency-gem)) to inherit from the version-controlled gem's YAML instead of the gihub.io file.

Instead of:
```
inherit_from:
  - https://shopify.github.io/ruby-style-guide/rubocop.yml
```
we'd use:
```
inherit_gem:
  ruby-style-guide: rubocop.yml
```

### Why?
The benefits are the same as any other gem - but the biggest win is projects can decide when to change versions and do so conveniently.

We would no longer need to commit the local cache file `.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml` and keep it up to date manually.

We could also formalize the associated rubocop version to our config by adding a gem dependency so that breaking changes to the YAML are reflected in a hard dependency.
eg. `spec.add_dependency 'rubocop', '~> 0.53'`

### Why not?
Dev galaxy or git submodules could be used for referencing this repo instead of a gem - and this isn't exactly pressing so maybe we just wait?

### TODO
- I just did the very basics of `bundle gem ...` and we'd need to do stuff for package cloud at the very least

### Thoughts?

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,19 @@
+The MIT License (MIT)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


rafaelfranca:
FYI, this is now true. This repository is a gem. Thank you for the inspiration @shawnfrench

volmer:
Damn I changed my mind so much after 1 year 😂
Take *that* ppl who said my strong opinions are strongly held!

rafaelfranca:
There is new tooling like dependabot helping with the goal that we didn't have at time.

shawnfrench:
Whoa amazing!! ❤️ 

Thanks for taking this to the finish line @rafaelfranca - I will make use of the new gem in our projects asap.

PR #143, opened 2020-02-26
Update docs to specify `require: false`

sambostock:
This updates the install instructions to include `require: false`. This is what `rubocop`, `rubocop-performance`, etc. do, so I assume it applies here too.

Not strictly related, but I wonder if we should indicate which Bundler groups the gem makes sense to be put in, since it isn't needed in production. I assume `:development, :test`, since it would be run locally, or in CI.


rafaelfranca:
Yeah, I think the group information would be interesting. Sorry didn't see that part until I merged. Can you open a new PR adding it?

sambostock:
No problem! 🤣 That was the fastest approval and merge I've seen in a while!

 Will do.

PR #144, opened 2020-02-26
Move recommended Gemfile entry into groups

sambostock:
Follow up to discussion in #143.

It should be recommended to include the gem in `:test` and `:development`, as it is required locally and in test/CI environments, but should not be installed in production.

I considered putting in a whole paragraph justifying why it should be in groups, and linking to the [Bundler groups documentation](https://bundler.io/guides/groups), but I realized that's probably overkill.


rafaelfranca:
Oh. I didn't thought about that. It is true. In a gem for example the entire Gemfile is only used on development and CI so there is no need for a group. Also, if you load this gem in production nothing will happen. The gem as no lib folder so no code is loaded.

PR #145, opened 2020-02-27
Add gem badge to README

larouxn:
Add gem badge to the README now that this style guide is exposed as a gem.
Screen Shot 2020-02-27 at 10 46 36

Follow-up to https://github.com/Shopify/ruby-style-guide/pull/142.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ permalink: '/index.html'
 ---
 
 # Ruby Style Guide
+[![Gem Version](https://badge.fury.io/rb/rubocop-shopify.svg)](https://badge.fury.io/rb/rubocop-shopify)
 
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request. Not sure what is the value that brings. I never liked those badges and the way I see that is just a way to those sites to collect information about projects. Can you explain what is the value we will get from this badge?

larouxn:
Valid point. I guess it's slightly faster to see the latest version but other than that I'm not sure adding a badge adds much value. Was more proposing we add it out of habit/tradition as most gems have badges like this.

larouxn:
Closing as this change is unnecessary.

PR #146, opened 2020-04-01
Fix Rubocop 0.81.0 compatibility

larouxn:
Rubocop v0.81.0 removed `Lint/EndInMethod` in favour of consolidating within `Style/EngBlock`. (https://github.com/rubocop-hq/rubocop/issues/7813)

Thus, I removed our `Lint/EndInMethod` rule. We already have `Style/EngBlock` specified here:
https://github.com/Shopify/ruby-style-guide/blob/b5e4bc98012c8717c52cb2c7d797ef2503094306/rubocop.yml#L635-L636

Example failure using `rubocop-shopify-1.0.0`:
> Error: The `Lint/EndInMethod` cop has been renamed to `Style/EndBlock`.
https://app.circleci.com/pipelines/github/larouxn/carrotgram/223/workflows/e803990d-8268-463c-8d91-ce3fe3d05b62/jobs/223

### DIFF
--- a/rubocop-shopify.gemspec
+++ b/rubocop-shopify.gemspec
@@ -20,5 +20,5 @@ Gem::Specification.new do |s|
     "source_code_uri"   => "https://github.com/Shopify/ruby-style-guide/tree/v#{s.version}",
   }
 
-  s.add_dependency "rubocop", ">= 0.78.0"
+  s.add_dependency "rubocop", ">= 0.81.0"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (1.0.0)
-      rubocop (>= 0.78.0)
+      rubocop (>= 0.81.0)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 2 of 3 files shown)


larouxn:
I suppose we'll need to release a v1.0.1 of this gem once this is merged. 🤔

volmer:
Yes we will. Also given that RuboCop ships breaking changes in minor versions we should limit the range of the compatible version of RuboCop to allow automatic patch upgrades only (e.g. `~> 0.81.0`)

larouxn:
Great points. I've updated the `gemspec` to require `"rubocop", "~> 0.81.0"` and ran `bundle install` to update the `Gemfile.lock` accordingly.

rubocop-shopify.gemspec
```diff
@@ -20,5 +20,5 @@ Gem::Specification.new do |s|
     "source_code_uri"   => "https://github.com/Shopify/ruby-style-guide/tree/v#{s.version}",
   }
 
-  s.add_dependency "rubocop", ">= 0.78.0"
+  s.add_dependency "rubocop", "~> 0.81.0"
```
rafaelfranca, line 23:
```suggestion
  s.add_dependency "rubocop", ">= 0.81.0"
```

If we don't allow people to install versions of rubocop that would break this gem we would never get PR like this anymore. Dependabot will not upgrade rubocop automatically in apps, and apps would not know that by upgrading rubocop  their CI fail.

This is without upper limit by design, no mistake. The upper limit doesn't protect anymore and break the automatic upgrade of gems.

PR #153, opened 2020-04-16
Rubocop 82 no longer supports 2.3, but macOS no longer ships 2.3

burke:


burke:
Thanks both for the quick :eyes:

PR #159, opened 2020-05-23
Speed up Rubocop in CI

sergey-alekseev:
## Before

From https://github.com/Shopify/ruby-style-guide/runs/700219213:  
>succeeded in 2m 8s

## After

From https://github.com/sergey-alekseev/ruby-style-guide/runs/701480507 (cold):  
>succeeded in 17s

From https://github.com/sergey-alekseev/ruby-style-guide/runs/701487963 (w/ cache):  
>succeeded in 16s

---
Replicated from [Rails](https://github.com/rails/rails/blob/100f7a7f643ae6766f406ccf1e049050199caddc/.github/workflows/rubocop.yml).

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -1,6 +1,6 @@
 name: Ruby
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -12,9 +12,15 @@ jobs:
     - name: Set up Ruby 2.7
       uses: actions/setup-ruby@v1
       with:
-        ruby-version: 2.7.1
+        ruby-version: 2.7
+    - name: Cache gems
+      uses: actions/cache@v1
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-rubocop-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-rubocop-
+    - name: Install gems
+      run: bundle install --jobs 4 --retry 3 --without pages --path vendor/bundle
     - name: RuboCop
-      run: |
-        gem install bundler
-        bundle install --jobs 4 --retry 3
-        bundle exec rubocop
+      run: bundle exec rubocop
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,3 +5,4 @@ AllCops:
   TargetRubyVersion: 2.7
   Exclude:
     - _site/**/*
+    - vendor/**/*


.github/workflows/ruby.yml
```diff
@@ -12,9 +12,18 @@ jobs:
...
+          ${{ runner.os }}-rubocop-
+    - name: Install gems
       run: |
-        gem install bundler
+        bundle config path vendor/bundle
+        bundle config set without 'pages'
         bundle install --jobs 4 --retry 3
-        bundle exec rubocop
+    - name: RuboCop
+      run: bundle exec rubocop --parallel
```
volmer, line 29: Why `--parallel`?

.rubocop.yml
```diff
@@ -5,3 +5,4 @@ AllCops:
   TargetRubyVersion: 2.7
   Exclude:
     - _site/**/*
+    - 'vendor/bundle/**/*'
```
volmer, line 8: If we need to exclude these then we can just `- vendor/**/*`

.github/workflows/ruby.yml
```diff
@@ -12,9 +12,18 @@ jobs:
...
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-rubocop-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-rubocop-
+    - name: Install gems
       run: |
-        gem install bundler
+        bundle config path vendor/bundle
+        bundle config set without 'pages'
```
volmer, line 26: Why is this config necessary?

sergey-alekseev:
>It's strange that there are no checks for your commit, though. Could you please verify?

I thought you did this on (some) purpose in 6c5edee3573421e5b487d1e3aec14b0cbde791c8. Do you want me to add it back in a separate commit as a fix?

.github/workflows/ruby.yml
sergey-alekseev: It's faster. E.g. look at [Rails](https://github.com/rails/rails/blob/100f7a7f643ae6766f406ccf1e049050199caddc/.github/workflows/rubocop.yml#L28) (again) and [Rubocop](https://github.com/rubocop-hq/rubocop/blob/v0.84.0/Rakefile#L21-L27).  

What is your concern?

.rubocop.yml
sergey-alekseev: We need to exclude these because gems are installed into this directory, see below. _vendor/bundle/_ is the only directory in _vendor/_, so I decided to be more specific and clear what exactly we exclude.

Do you think there is any reason to exclude the whole _vendor/_ at the moment?

.github/workflows/ruby.yml
sergey-alekseev: To get the major speed gain.

From [the commit](https://github.com/rails/rails/commit/100f7a7f643ae6766f406ccf1e049050199caddc) mentioned in the description:  
>Unfortunately it's not possible to tell Bundler to only install gems
from a single group, so we have to tell it not to install every other
group instead.

Any proposal to improve this?

volmer:
> I thought you did this on (some) purpose in 6c5edee. Do you want me to add it back in a separate commit as a fix?

I disabled the action for the `pull_request` event because I assumed the `push` event would cover all cases. Your PR comes from a fork though, so there was no `push` event in this repo, that's why the check is missing. Yes, please add the pull request event back if possible 🙏

.github/workflows/ruby.yml
volmer: I understand the optimization but we only have a couple of `.rb` in this repo, and they are very small. We don't need to run cops in parallel, it's not gonna give us any significant boost in the check speed.

.rubocop.yml
volmer: Yes, we should ignore the whole `vendor`. There are no files for RuboCop to analyze under `vendor`.

.github/workflows/ruby.yml
volmer: I understand we don't need to install gems under the `pages` group, but what about specifying the `--without=pages` flag to `bundle install` instead for simplicity?

.github/workflows/ruby.yml
sergey-alekseev: It's just a generally faster approach. Of course, we don't care about `time rubocop` in this repo at the moment. OK, changed.

.rubocop.yml
sergey-alekseev: There is even no `vendor/` directory yet. OK, changed.

.github/workflows/ruby.yml
sergey-alekseev: I see no difference in this case. OK, changed.  

From https://github.com/Shopify/ruby-style-guide/pull/159/checks?check_run_id=733139479:  
```
Run bundle install --jobs 4 --retry 3 --without pages --path vendor/bundle
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag
[DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'pages'`, and stop using this flag
```

sergey-alekseev:
>Yes, please add the pull request event back if possible 🙏

Done in 391dc6a1079c9cdf20016f67ffaf7cc26f750083.

sergey-alekseev:
@volmer I addressed all your comments which you posted so far.

sergey-alekseev:
Thanks for the review! @volmer  

May be useful to go over the resolved comments and particularly deprecated options in https://github.com/Shopify/ruby-style-guide/pull/159#discussion_r434270725. @rafaelfranca

volmer:
Thank you @sergey-alekseev ! 🙏

PR #164, opened 2020-06-02
Enable Lint/DeprecatedOpenSSLConstant cop

bdewater:
This will prevent warnings coming from the changes in https://github.com/ruby/openssl/pull/366

### DIFF
--- a/rubocop-shopify.gemspec
+++ b/rubocop-shopify.gemspec
@@ -21,5 +21,5 @@ Gem::Specification.new do |s|
     "allowed_push_host" => "https://rubygems.org",
   }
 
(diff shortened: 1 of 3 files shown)


rubocop-shopify.gemspec
```diff
@@ -21,5 +21,5 @@ Gem::Specification.new do |s|
     "allowed_push_host" => "https://rubygems.org",
   }
 
-  s.add_dependency("rubocop", ">= 0.82", "< 0.86")
+  s.add_dependency("rubocop", ">= 0.84", "< 0.86")
```
volmer, line 24: I think we can enforce 0.85.0 and later already.

```suggestion
  s.add_dependency("rubocop", "~> 0.85.0")
```

PR #167, opened 2020-06-11
address discrimination when naming things

kevinodotnet:
Updates the style guide to guide developers to avoid nomenclature that has (or could be interpreted to have) discriminatory origins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,7 +451,7 @@ For more information about inheriting configuration from a gem please check
 
 * Avoid magic numbers. Use a constant and give it a useful name.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -451,6 +451,7 @@ For more information about inheriting configuration from a gem please check
 
 * Avoid magic numbers. Use a constant and give it a useful name.
 
+* Avoid nomenclature that has (or could be interpretted to have) discriminatory origins.
```
andyw8, line 454: Since this would apply to all programming languages used at Shopify, is there any other shared guide where this could be covered?

(also, typo in 'interpretted').

PR #168, opened 2020-06-15
Add 'Avoid the usage of class instance variables'

bdewater:
Per the discussion on Slack earlier today, where we reviewed internal code where seemingly the authors were not aware of the `self << class` and used class instance variables instead of instance variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,6 +584,46 @@ For more information about inheriting configuration from a gem please check
 * Avoid the usage of class (`@@`) variables due to their "nasty" behavior in
   inheritance.
 
+* Avoid using instance variables on singleton objects (for example, classes
+  and modules), as they might cause thread-safety issues.
+
+  ~~~ ruby
+  # bad
+  class Processor
+    class << self
+      attr_reader :thing
+
+      def call(thing, action)
+        @thing = thing
+        thing.prepare_for_action
+        process(action)
+      end
+
+      private
+
+      def process(action)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -584,6 +584,68 @@ For more information about inheriting configuration from a gem please check
 * Avoid the usage of class (`@@`) variables due to their "nasty" behavior in
   inheritance.
 
+* Avoid the usage of class instance variables unless you really need them
```
rafaelfranca, line 587:
```suggestion
* Avoid the usage of instance variables on singletons unless you really need them
```

bdewater: Thanks for the suggestion, I slightly altered the entire text to incorporate it.

gmalette:
To be candid, I don't think this directive belongs in a *style* guide. It's a software development issue, not a style one.

README.md
```diff
@@ -584,6 +584,68 @@ For more information about inheriting configuration from a gem please check
...
+      process(action)
+    end
+
+    private
+
+    def process(action)
+      Actions.for(action).perform_with(thing)
+    end
+  end
+  ~~~
```
volmer, line 647: This is a pretty long example for a simple rule. Could we maybe rethink these code snippets and exemplify the rule with something a bit simpler?

bdewater:
I'm back from some time off :) @volmer thanks for your suggestion, I've trimmed the fat.

PR #174, opened 2020-08-05
rubocop 0.89 compatibility

exterm:
Sadly, 0.89 renames some cops (Lint/MissingSuper, Lint/BinaryOperatorWithIdenticalOperands).

0.88 will not accept the new names, while 0.89 will not accept the old names. I would have expected a grace period, but alas.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (1.0.4)
-      rubocop (>= 0.85, < 0.89)
+      rubocop (~> 0.89.0)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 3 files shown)


rubocop-shopify.gemspec
```diff
@@ -21,5 +21,5 @@ Gem::Specification.new do |s|
     "allowed_push_host" => "https://rubygems.org",
   }
 
-  s.add_dependency("rubocop", ">= 0.85", "< 0.89")
+  s.add_dependency("rubocop", "0.89")
```
volmer, line 24:
```suggestion
  s.add_dependency("rubocop", "~> 0.89.0")
```

We need to allow patches for this version.

Gemfile.lock
```diff
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (1.0.4)
-      rubocop (>= 0.85, < 0.89)
+      rubocop (= 0.89)
```
volmer, line 5: You might need to run `bundle install` to fix this.

exterm: I don't see anything wrong?

exterm: ah! Got it.

PR #186, opened 2020-09-14
Table of Contents

### LINKED ISSUE #74: Style guide style improvements
Not a huge deal, but I thought I'd throw this out there.

I really like the way AirBnB does their [Javascript Style Guide](https://github.com/airbnb/javascript), for the following reasons:

- Table of Contents 📄
- Linking to specific sections 👈
- Examples for almost everything, often with 😇 good vs 👿 bad
- "Why"s on many style rules 🤔 

In my opinion, these make the style guide much more approachable to developers new to a language, as well as making the rules easier to share with others/refer to. I like having a "why" too, because rules make so much more sense when they aren't arbitrary.

### LINKED ISSUE #187: Add a Table of Contents
Extracted from #74, having a Table of Contents in the beginning of the document would make it easier to read.

cc @sambostock

volmer:
This change adds a "Table of Contents" to the beginning of the document to help developers navigate through the sections of the style guide. It also adds anchors to each section so you can link to specific subsets of rules.

As an aside, by working on this I realize this style guide became quite long. I might take some time to review and remove redundancies, deprecations, and others. Expect another PR soon!

Fixes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,21 @@ For more information about inheriting configuration from a gem please check
 [RuboCop's
 documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-configuration-from-a-dependency-gem).
 
+## Table of Contents
+
+* [General](#general)
+* [Formatting](#formatting)
+* [Syntax](#syntax)
+* [Naming](#naming)
+* [Classes and Modules](#classes-and-modules)
+* [Exceptions](#exceptions)
+* [Collections](#collections)
+* [Strings](#strings)
+* [Regular Expressions](#regular-expressions)
+* [Percent Literals](#percent-literals)
+* [Testing](#testing)
+* [The Rest](#the-rest)
+
 ## General
 
 * Make all lines of your methods operate on the same level of abstraction.
@@ -70,7 +85,6 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Be consistent.
 
 * Use common sense.
-
 
 ## Formatting
 
@@ -212,7 +226,6 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
     arg_2,
   )
   ~~~
-
 
 ## Syntax
 
@@ -490,8 +503,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * For a good discussion on the costs and benefits of comments, see
   [http://c2.com/cgi/wiki?CommentCostsAndBenefits](http://c2.com/cgi/wiki?CommentCostsAndBenefits).
 
-
-## Classes & Modules
+## Classes and Modules
 
 * Prefer modules to classes with only class methods. Classes should be used only
   when it makes sense to create instances out of them.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,22 @@ For more information about inheriting configuration from a gem please check
...
+* [Classes and Modules](#classes-and-modules)
+* [Exceptions](#exceptions)
+* [Collections](#collections)
+* [Strings](#strings)
+* [Regular Expressions](#regular-expressions)
+* [Percent Literals](#percent-literals)
+* [Testing](#testing)
+* [The Rest](#the-rest)
+
+## General <a name="general"></a>
```
lugray, line 63: No need to add anchors, markdown headers have them created on the site and in github automatically:
 - https://shopify.github.io/ruby-style-guide/#general
 - https://github.com/Shopify/ruby-style-guide#general

sambostock (commented):
What a blast from the past! Thanks for addressing such an old issue!

As @lugray mentioned, we can already link to section headers, so we should probably use the out of the box anchors, unless we need to customize them or something. The table of contents is a nice addition though.

What I think would be most impactful would be the ability to link to a **specific rule**. For example, I can link directly to [the rule on implicit return in arrow functions](https://github.com/airbnb/javascript#arrows--implicit-return) in [AirBnB's JavaScript Styleguide](https://github.com/airbnb/javascript).

README.md
volmer: TIL! Thanks

volmer:
> What I think would be most impactful would be the ability to link to a specific rule

That would be very nice. I'm afraid it would be only possible if we manually add anchors to each of our 100+ rules. Not sure if the benefit justifies the effort. It would be an addition unrelated to this PR in any case.

sambostock:
That's fair, and I'm glad to approve the addition of this addition of a table of contents, but I don't think this `Fixes #74` completely on it's own.

volmer:
> I don't think this Fixes #74 completely on it's own.

Would you like me to edit #74 to reflect that it's only about rule anchors from now on? If you prefer I can open a new issue just for that.

sambostock:
Hmm... #74 lists four separate points:

> - Table of Contents 📄

This PR nicely addresses this.

> - Linking to specific sections 👈

_As worded_ this was already possible, but ideally we would be able to link to individual rules (which I think is what I originally meant to convey in that issue).

> - Examples for almost everything, often with 😇 good vs 👿 bad
> - "Why"s on many style rules 🤔

These are out of scope of this PR.

Do you think they should be four separate issues, discussed and addressed separately, or is this PR just the first of many to address all aspects of #74?

volmer:
I’d rather have each of these as separate issues.

lugray:
If we made each rule H3, I think we'd get rule anchors for free.  CSS could style the H3 as bullets.

sambostock:
I think it would make sense to use an approach that works on the GitHub README as well. Making the entire rule an `<h3>` probably wouldn't work well. Alternatives would include having a title for each rule, or keeping bullet points, but putting an actor at the front (as AirBnB does with anchored rule numbers).

volmer:
Ok I broke the original issue into more specific ones. This PR now tackles only #187.

sambostock (approved):
:shipit:

Edouard-chin (approved):

PR #191, opened 2020-09-15
Enable ImplicitRuntimeError

catherinetcai:
https://github.com/Shopify/services-db/issues/6027

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -1019,3 +1019,6 @@ Lint/OrderedMagicComments:
 
 Lint/DeprecatedOpenSSLConstant:
   Enabled: true
+
+Style/ImplicitRuntimeError:
+  Enabled: true


ryanwilsonperkin (approved):
🎉 Woooo

volmer (changes requested):
Please note that this is a public repository. It would be great if you could provide full context about the change you are proposing in the PR description. The issue you reference is located in a private repository.

Regarding the cop you are trying to enable, the rule it enforces contradicts this Style Guide, which states that you should keep `RuntimeError` implicit:

> Don't specify `RuntimeError` explicitly in the two argument version of `raise`.
>
> ```ruby
> # bad
> raise RuntimeError, 'message'
> 
> # good - signals a RuntimeError by default
> raise 'message'
> ```

https://shopify.github.io/ruby-style-guide/#exceptions

catherinetcai:
Ah, my bad. I'll close and update accordingly. Cheers.

PR #193, opened 2020-09-17
Make wording and tone more consistent

volmer:
In order to improve readability, the Style Guide should use a more consistent wording when stating its rules. 

In this change I tried to reword rules using always the imperative tense, which is more assertive and leaves less space for doubt. So I replaced cases such as:

> Using the return value of = in a conditional expression is okay if wrapped in parenthesis.

With:

> Wrap assignment in parentheses when using its return value in a conditional statement.

I also tried to make wording in rules that say "don't do this" more consistent. We were phrasing these case using expressions such as "do not" , "never", "refrain from", which for me they all mean "avoid doing this unless you have a very good reason to break the rules". So I changed all cases from a negative tone to an assertive tone with "avoid". So I switched from:

> Don't use block comments.

to

> Avoid block comments.

I also fixed punctuation and reorganized examples to include any clarification necessary.

Please keep in mind that I'm still learning to write in English and that I'm still pretty bad at it, so any feedback/correction is much appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,18 +67,15 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Code in a functional way. Avoid mutation (side effects) when you can.
 
-* Do not program defensively (see
-  [http://www.erlang.se/doc/programming_rules.shtml#HDR11](http://www.erlang.se/doc/programming_rules.shtml#HDR11)).
+* [Avoid defensive programming](http://www.erlang.se/doc/programming_rules.shtml#HDR11)
 
-* Do not mutate arguments unless that is the purpose of the method.
+* Avoid mutating arguments.
 
-* Do not mess around in / monkeypatch core classes when writing libraries.
+* Avoid monkeypatching.
 
 * Keep the code simple.
 
-* Don't overdesign.
-
-* Don't underdesign.
+* Do the right amount of design, not too little but also not too much.
 
 * Avoid bugs.
 
@@ -94,17 +91,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Use Unix-style line endings.
 
-* Don't use `;` to separate statements and expressions. As a corollary - use one
+* Avoid using `;` to separate statements and expressions. Use one
   expression per line.
 
 * Use spaces around operators, after commas, colons and semicolons, around `{`
   and before `}`.
 
-* No spaces after `(`, `[` and before `]`, `)`.
+* Avoid spaces after `(`, `[` and before `]`, `)`.
 
-* No space after the `!` operator.
+* Avoid space after the `!` operator.
 
-* No space inside range literals.
+* Avoid space inside range literals.
 
 * Indent `when` as deep as the `case` line.
 
@@ -138,8 +135,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Use spaces around the `=` operator when assigning default values to method
   parameters.
 
-* Avoid line continuation `\` where not required. In practice, avoid using line
-  continuations for anything but string concatenation.
+* Avoid line continuation `\` where not required.
 
 * Align the parameters of a method call, if they span more than one line, with
   one level of indentation relative to the start of the line with the method
@@ -197,7 +193,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * End each file with a newline.
 
-* Don't use block comments:
+* Avoid block comments:
 
   ~~~ ruby
   # bad
@@ -211,7 +207,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   # another comment line
   ~~~
 
-* Closing method call brace must be on the line after the last argument when
+* Place the closing method call brace on the line after the last argument when
   opening brace is on a separate line from the first argument.
 
   ~~~ ruby
@@ -230,7 +226,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 ## Syntax
 
 * Use `::` only to reference constants (this includes classes and modules) and
-  constructors (like `Array()` or `Nokogiri::HTML()`). Do not use `::` for
+  constructors (like `Array()` or `Nokogiri::HTML()`). Avoid `::` for
   regular method invocation.
 
 * Avoid using `::` for defining class and modules, or for inheritance, since
@@ -259,9 +255,9 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Use def with parentheses when there are parameters. Omit the parentheses when
   the method doesn't accept any parameters.
 
-* Never use `for`, unless you know exactly why.
+* Avoid `for`.
 
-* Never use `then`.
+* Avoid `then`.
 
 * Favour the ternary operator(`?:`) over `if/then/else/end` constructs.
 
@@ -277,18 +273,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   ternary operators must not be nested. Prefer if/else constructs in these
   cases.
 
+* Avoid multiline `?:` (the ternary operator); use `if/unless` instead.
+
 * Use `when x then ...` for one-line cases.
 
 * Use `!` instead of `not`.
 
-* Prefer `&&`/`||` over `and`/`or`. [More info on `and/or` for control
-  flow](http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/).
-
-* Avoid multiline `?:` (the ternary operator); use `if/unless` instead.
+* Prefer `&&`/`||` over `and`/`or`.
 
 * Favour `unless` over `if` for negative conditions.
 
-* Do not use `unless` with `else`. Rewrite these with the positive case first.
+* Avoid `unless` with `else`. Rewrite these with the positive case first.
 
 * Use parentheses around the arguments of method invocations. Omit parentheses
   when not providing arguments. Also omit parentheses when the invocation is
@@ -320,22 +315,6 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
     * `raise`
     * `puts`
 
-* Use class methods instead of a rails scope with a multi-line lambda
-
-  ~~~ ruby
-  # bad
-  scope(:pending, -> do
-    ...
-    ...
-  end)
-
-  # good
-  def self.pending
-    ...
-    ...
-  end
-  ~~~
-
 * Omit the outer braces around an implicit options hash.
 
 * Use the proc invocation shorthand when the invoked method is the only
@@ -349,27 +328,39 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   names.map(&:upcase)
   ~~~
 
-* Prefer `{...}` over `do...end` for single-line blocks. Avoid using `{...}` for
-  multi-line blocks. Always use `do...end` for "control flow" and "method
-  definitions" (e.g. in Rakefiles and certain DSLs). Avoid `do...end` when
-  chaining.
+* Prefer `{...}` over `do...end` for single-line blocks.
 
-* Avoid `return` where not required for control flow.
+* Prefer `do..end` over `{...}` for multi-line blocks.
 
-* Avoid `self` where not required (it is only required when calling a self write
-  accessor).
+* Omit `return` where possible.
 
-* Using the return value of = in a conditional expression is okay if wrapped in
-  parenthesis
+* Omit `self` where possible.
 
   ~~~ ruby
-  if (v = /foo/.match(string))
+  # bad
+  self.my_method
+
+  # good
+  my_method
+
+  # also good
+  attr_writer :name
+
+  def my_method
+    self.name = 'Rafael' # `self` is neeeded to reference the attribute writer.
+  end
+  ~~~
+
+* Wrap assignment in parentheses when using its return value in a conditional
+  statement.
+
+  ~~~ ruby
+  if (value = /foo/.match(string))
   ~~~
 
 * Use `||=` to initialize variables only if they're not already initialized.
 
-* Don't use `||=` to initialize boolean variables (consider what would happen if
-  the current value happened to be `false`).
+* Avoid using `||=` to initialize boolean variables.
 
   ~~~ ruby
   # bad - would set enabled to true even if it was false
@@ -382,9 +373,9 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   @enabled = true unless defined?(@enabled)
   ~~~
 
-* Do not put a space between a method name and the opening parenthesis.
+* Avoid spaces between a method name and the opening parenthesis.
 
-* Use the new lambda literal syntax.
+* Prefer the lambda literal syntax over `lambda`.
 
   ~~~ ruby
   # bad
@@ -408,8 +399,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Prefer `proc` over `Proc.new`.
 
-* Prefix with `_` unused block parameters and local variables. It's also
-  acceptable to use just `_`.
+* Prefix unused block parameters with `_`. It's also acceptable to use just `_`.
 
 * Prefer a guard clause when you can assert invalid data. A guard clause is a
   conditional statement at the top of a function that bails out as soon as it
@@ -437,31 +427,27 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   end
   ~~~
 
-* Avoid hashes-as-optional-parameters in general. Does the method do too much?
-
 * Prefer keyword arguments over options hash.
 
 * Prefer `map` over `collect`, `find` over `detect`, `select` over `find_all`,
   `size` over `length`.
 
-* Prefer `Time` over `DateTime` since it supports proper time zones instead of
-  UTC offsets. [More
-  info](https://gist.github.com/pixeltrix/e2298822dd89d854444b).
+* Prefer `Time` over `DateTime`.
 
 * Prefer `Time.iso8601(foo)` instead of `Time.parse(foo)` when expecting ISO8601
   formatted time strings like `"2018-03-20T11:16:39-04:00"`.
 
 ## Naming
 
-* Use `snake_case` for symbols, methods and variables.
+* Use `snake_case` for symbols, methods, and variables.
 
-* Use `CamelCase` for classes and modules (keep acronyms like HTTP, RFC, XML
-  uppercase).
+* Use `CamelCase` for classes and modules, but keep acronyms like HTTP, RFC, XML
+  uppercase.
 
 * Use `snake_case` for naming files and directories, e.g. `hello_world.rb`.
 
-* Aim to have just a single class/module per source file. Name the file name as
-  the class/module, but replacing `CamelCase` with `snake_case`.
+* Define a single class or module per source file. Name the file name as the
+  class or module, but replacing `CamelCase` with `snake_case`.
 
 * Use `SCREAMING_SNAKE_CASE` for other constants.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,8 +137,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Use spaces around the `=` operator when assigning default values to method
   parameters.
 
-* Avoid line continuation `\` where not required. In practice, avoid using line
-  continuations for anything but string concatenation.
+* Avoid line continuation `\` where not required.
```
volmer, line 140: Removed the last sentence since it doesn't add anything to the rule imo.

README.md
```diff
@@ -277,18 +275,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
   cases.
 
+* Avoid multiline `?:` (the ternary operator); use `if/unless` instead.
+
 * Use `when x then ...` for one-line cases.
 
 * Use `!` instead of `not`.
 
-* Prefer `&&`/`||` over `and`/`or`. [More info on `and/or` for control
-  flow](http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/).
```
volmer, line 285: Removed the link as I find it unnecessary.

README.md
```diff
@@ -277,18 +275,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   ternary operators must not be nested. Prefer if/else constructs in these
   cases.
 
+* Avoid multiline `?:` (the ternary operator); use `if/unless` instead.
```
volmer, line 278: I'm just moving this rule to be closer to the one regarding ternary operators.

README.md
```diff
@@ -277,18 +275,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
+
 * Use `when x then ...` for one-line cases.
 
 * Use `!` instead of `not`.
 
-* Prefer `&&`/`||` over `and`/`or`. [More info on `and/or` for control
-  flow](http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/).
-
-* Avoid multiline `?:` (the ternary operator); use `if/unless` instead.
```
volmer, line 287: I'm just moving this rule to be closer to the one regarding ternary operators.

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
-* Focus on **why** your code is the way it is if this is not obvious, not
-  **how** your code works.
-
-* Avoid superfluous comments. If they are about how your code works, should you
-  clarify your code instead?
-
-* For a good discussion on the costs and benefits of comments, see
-  [http://c2.com/cgi/wiki?CommentCostsAndBenefits](http://c2.com/cgi/wiki?CommentCostsAndBenefits).
+* Avoid superfluous comments. Focus on **why** the code is the way it is if
```
volmer, line 493: Merged these three rules into a simpler single version.

README.md
```diff
@@ -791,8 +777,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   hash = {}
   ~~~
 
-* Prefer the literal array syntax to `%w`, except when it reads substantially
-  more clearly in context.
```
volmer, line 795: "except when..." is implicit not just for the rule but for pretty much everything in this style guide.

README.md
```diff
@@ -802,66 +787,53 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
-  # encouraged
+  # good
   {
     foo: :bar,
     baz: :toto,
   }
   ~~~
 
-* Prefer the literal array syntax to `%i`.
```
volmer, line 822: Merge this with the rule at the top.

README.md
```diff
@@ -1127,18 +1093,12 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
-* A good test case consists of three parts:
-    1. Setup of the environment
-    2. The action that is the subject of the test
-    3. Asserting that the action did what you expect it do to.
+* Limit each test case to cover a single aspect of your code.
 
-  Consider separating these parts by a newline for readability, especially when
-  your environment setup is complicated and you want to run multiple assertions
-  afterwards.
+* Organize the setup, action, and assertion sectinos of the test case with empty
```
Edouard-chin, line 1100:
```suggestion
* Organize the setup, action, and assertion sections of the test case with empty
```

README.md
```diff
@@ -578,13 +567,11 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   end
   ~~~
 
-* When designing class hierarchies make sure that they conform to the [Liskov
-  Substitution
-  Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).
+* Respect the [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle)
+  When designing class hierarchies.
```
Edouard-chin, line 571:
```suggestion
  when designing class hierarchies.
```

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * When defining binary operators, name the parameter `other`(`<<` and `[]` are
   exceptions to the rule, since their semantics are different).
 
-* The names of predicate methods (methods that return a boolean value) should
-  end in a question mark (i.e. `Array#empty?`). Methods that don't return a
-  boolean, shouldn't end in a question mark.
+* Name predicate methods with a question mark. Predicate methods are methods
```
Edouard-chin, line 462: I'd replace the `question mark` to  `?` . In other part of this document, characters such as slash, brackets etc... aren't written literally

Edouard-chin (approved):

README.md
```diff
@@ -67,18 +67,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Code in a functional way. Avoid mutation (side effects) when you can.
 
-* Do not program defensively (see
-  [http://www.erlang.se/doc/programming_rules.shtml#HDR11](http://www.erlang.se/doc/programming_rules.shtml#HDR11)).
+* [Avoid defensive programming](http://www.erlang.se/doc/programming_rules.shtml#HDR11)
 
-* Do not mutate arguments unless that is the purpose of the method.
+* Keep method arguments unchanged.
```
sambostock, line 72: I think this could be phrased using `avoid`, as you've done in other places
```suggestion
* Avoid mutating arguments, unless that is the purpose of the method.
```

README.md
```diff
@@ -349,27 +330,39 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
   ~~~ ruby
-  if (v = /foo/.match(string))
+  if (value = /foo/.match(string))
   ~~~
 
 * Use `||=` to initialize variables only if they're not already initialized.
 
-* Don't use `||=` to initialize boolean variables (consider what would happen if
-  the current value happened to be `false`).
+* Avoid using `||=` to initialize boolean variables.
```
sambostock, line 365: Might be worth extending this from just booleans to also values that could be `nil`.

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
 * Avoid nomenclature that has (or could be interpreted to have) discriminatory
   origins.
 
 ## Comments
 
-* Good comments focus on the reader of the code, by helping them understand the
-  code. The reader may not have the same understanding, experience and knowledge
-  as you. As a writer, take this into account.
+* When writing comments account for the fact that readers might not have the
```
sambostock, line 486:
```suggestion
* Include relevant context in comments, as readers might be missing it.
```

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
+  same context as you do.
 
-* A big problem with comments is that they can get out of sync with the code
-  easily. When refactoring code, refactor the surrounding comments as well.
+* Keep surrounding comments in sync when refactoring code.
```
sambostock, line 489: Arguably, refactoring is the one case where the comments might not need to change. 😅 

```suggestion
* Keep comments in sync with code.
```

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
-* Write good copy, and use proper capitalization and punctuation.
+* Write good copy, and use proper capitalization and punctuation in comments.
```
sambostock, line 491: Personally, I'm not sure what "good copy" means in this context.

```suggestion
* Write comments using proper capitalization and punctuation.
```

README.md
```diff
@@ -802,66 +787,53 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
   # good
-  { :a => 1, 'b' => 2 }
+  { a: 1, b: 2 }
   ~~~
 
-* Use `Hash#key?` instead of `Hash#has_key?` and `Hash#value?` instead of
-  `Hash#has_value?`. As noted
-  [here](http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/43765) by
-  Matz, the longer forms are considered deprecated.
+* Avoid mixing the Ruby 1.9 hash syntax with hash rockets in the same hash
```
sambostock, line 821: We could probably merge this with the previous rule by saying something like this and combining the examples.

```markdown
* Use shorthand hash literal syntax when **all** keys are symbols
```

I also don't think referring to Ruby 1.9 is very clear, as most developers at Shopify will have started writing Ruby well into 2.x.

README.md
```diff
@@ -1057,8 +1028,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   /(?:first|second)/
   ~~~
 
-* Don't use the cryptic Perl-legacy variables denoting last regexp group matches
-  (`$1`, `$2`, etc). Use `Regexp#match` instead.
+* Prefer `Regexp#match` over Perl-legacy variables to capture group matches.
```
sambostock, line 1031: Keeping

> (`$1`, `$2`, etc)

would make it clearer what this refers to for people who haven't used Perl.

README.md
```diff
@@ -1127,18 +1093,12 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
-* A good test case consists of three parts:
-    1. Setup of the environment
-    2. The action that is the subject of the test
-    3. Asserting that the action did what you expect it do to.
+* Limit each test case to cover a single aspect of your code.
 
-  Consider separating these parts by a newline for readability, especially when
-  your environment setup is complicated and you want to run multiple assertions
-  afterwards.
+* Organize the setup, action, and assertion sections of the test case with empty
```
sambostock, line 1100:
```suggestion
* Organize the setup, action, and assertion sections of the test case into paragraphs separated by empty
```

README.md
```diff
@@ -1196,13 +1156,12 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
-* Never use a `!` at the end of a method name if you have no equivalent method
-  without the bang. Methods are expected to change internal object state; you
-  don't need a bang for that. Bangs are to mark a more dangerous version of a
-  method, e.g. `save` returns a `bool` in ActiveRecord, whereas `save!` will
-  throw an exception on failure.
+* Avoid ending method names with `!` when there is no equivalent method without
+  the bang. Bangs are to mark a more dangerous version of a method, e.g. `save`
+  returns a `bool` in ActiveRecord, whereas `save!` will throw an exception on
+  failure.
```
sambostock, line 1164: Might be worth mentioning the mutation use case as well (e.g. `map` vs `map!`).

Also "`bool`" should probably just be "boolean"

sambostock (commented):
Couple suggestions, but overall I think these are good changes 👍

README.md
PragTob: Not sure, I think it's interesting. Like I vaguely remember that there was some weird difference esp. folks new to Ruby might be like "but why?"

README.md
```diff
@@ -320,22 +317,6 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
     * `raise`
     * `puts`
 
-* Use class methods instead of a rails scope with a multi-line lambda
```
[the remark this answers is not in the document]
PragTob: Do we have a rails style guide somewhere where this is metnioned?

README.md
```diff
@@ -382,9 +375,9 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   @enabled = true unless defined?(@enabled)
   ~~~
 
-* Do not put a space between a method name and the opening parenthesis.
+* Avoid spaces between a method name and the opening parenthesis.
```
PragTob, line 378: I think this should also be kept as `do not`

README.md
```diff
@@ -437,31 +429,27 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   end
   ~~~
 
-* Avoid hashes-as-optional-parameters in general. Does the method do too much?
```
PragTob, line 440: Did this move somewhere that I didn't see or do we want to get rid of it?

README.md
```diff
@@ -437,31 +429,27 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
-
 * Prefer keyword arguments over options hash.
 
 * Prefer `map` over `collect`, `find` over `detect`, `select` over `find_all`,
   `size` over `length`.
 
-* Prefer `Time` over `DateTime` since it supports proper time zones instead of
-  UTC offsets. [More
-  info](https://gist.github.com/pixeltrix/e2298822dd89d854444b).
+* Prefer `Time` over `DateTime`.
```
PragTob, line 437: Imo saying TZ vs offsets provides a nice and short rational

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * When defining binary operators, name the parameter `other`(`<<` and `[]` are
   exceptions to the rule, since their semantics are different).
 
-* The names of predicate methods (methods that return a boolean value) should
-  end in a question mark (i.e. `Array#empty?`). Methods that don't return a
-  boolean, shouldn't end in a question mark.
+* Name predicate methods with a `?`. Predicate methods are methods that return a
+  boolean value.
```
PragTob, line 463: Imo it's also an imortant property that they don't have a side effect but iirc that's somewhat contentious

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
-* Method names should not be prefixed with `is_`. E.g. prefer `empty?` over
-  `is_empty?`.
+  # good
+  def empty?
+  end
+  ~~~
 
-* Avoid magic numbers. Use a constant and give it a useful name.
+* Avoid magic numbers. Use a constant and give it a meaningful name.
```
PragTob, line 479: 👍

README.md
```diff
@@ -471,45 +459,46 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 * Avoid nomenclature that has (or could be interpreted to have) discriminatory
   origins.
 
 ## Comments
 
-* Good comments focus on the reader of the code, by helping them understand the
-  code. The reader may not have the same understanding, experience and knowledge
-  as you. As a writer, take this into account.
+* When writing comments account for the fact that readers might not have the
+  same context as you do.
```
PragTob, line 487: I don't think it's up yet, but soon there will be a new section in the development handbook dealing with source code comments (the google doc is already completely reviewed)

README.md
```diff
@@ -614,15 +601,13 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   end
   ~~~
 
-* Avoid the use of `attr`. Use `attr_reader` and `attr_accessor` instead.
+* Prefer `attr_reader` and `attr_accessor` over `attr`.
 
-* Avoid the usage of class (`@@`) variables due to their "nasty" behavior in
-  inheritance.
+* Avoid class (`@@`) variables.
```
PragTob, line 606: here I also like the little reference to the reason, esp. people for new to ruby it might be nice so to see it's not just arbitrary rules but there's actual danger in using these.

README.md
```diff
@@ -802,66 +787,53 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   STATES = ['draft', 'open', 'closed']
   ~~~
 
-* Usage of trailing comma in multi-line collection literals is encouraged. It
-  makes diffs smaller and more meaningful.
+* Append a trailing comma in multi-line collection literals.
```
PragTob, line 790: the change is pretty hard from "it is encouraged" to "do it" maybe we could add a prefer or something. Unless ofc we want that ALL the time.

README.md
```diff
@@ -976,18 +945,17 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   puts "$global = #{$global}"
   ~~~
 
-* Don't use `Object#to_s` on interpolated objects. It's invoked on them
-  automatically.
+* Avoid `Object#to_s` on interpolated objects.
```
PragTob, line 948: Imo that's another `don't` case because it's directly what it does, I can't imagine needing this in any scenario, but maybe it's my imagination

README.md
```diff
@@ -1127,18 +1093,12 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Treat test code like any other code you write. This means: keep readability,
   maintainability, complexity, etc. in mind.
 
-* Minitest is the preferred test framework.
-
-* A test case should only test a single aspect of your code.
+* Use Minitest as test framework.
```
PragTob, line 1096: It's missing `the` right? I wish this would be `prefer` but I understand 👼

README.md
```diff
@@ -1196,13 +1156,12 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Avoid needless metaprogramming.
 
-* Never start a method with `get_`.
+* Avoid starting method names with `get_`.
```
PragTob, line 1159: feels like this could be moved up to where we discuss predicate/? methods

README.md
```diff
@@ -1196,13 +1156,12 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
 
-* Never use a `!` at the end of a method name if you have no equivalent method
-  without the bang. Methods are expected to change internal object state; you
-  don't need a bang for that. Bangs are to mark a more dangerous version of a
-  method, e.g. `save` returns a `bool` in ActiveRecord, whereas `save!` will
-  throw an exception on failure.
+* Avoid ending method names with `!` when there is no equivalent method without
```
PragTob, line 1161: also would probably move it up to the `?` section

PragTob (commented):
Looking good!

Thanks for doing this work :)

Added some comments, I realized that I think that some rules should always be applied. Especially when it's some really basic layouting that doesn't have any real difference and rubocop would complain about it anyhow. So there I like the old _don't_ usage. Plus things that are either so "broken" they shouldn't be used or have a complete functional equivalent (`::` to call methods...)

WhatsApp Image 2020-09-18 at 07 53 37(3)

README.md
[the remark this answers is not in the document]
volmer: We don't have a Rails Style Guide yet.

README.md
volmer: Applied! Except the "unless that is the purpose of the method" part since we shouldn't be appending "except when it makes sense to break the rule" because that's implicit.

README.md
volmer: I used the code sample to exemplify `$1`. We should always prefer to clarify a rule with a code sample with necessary (instead of using extra sentences or parentheses).

README.md
volmer: Replaced `bool` with "boolean". I'm not mentioning the mutation use case because that's the one we are encouraging people to avoid.

README.md
volmer: With this change I'm removing such bits of rationale from rules to make them more concise. The fact that `Time` supports time zones is not the reason why people should follow the rule. It's because it makes code more consistent (like all the other rules).

README.md
volmer: Although I very much agree with you I think mentioning that would mean introducing new rules to the style guide, which I am avoiding in this PR.

README.md
volmer: Good to know!

README.md
volmer: Same as above. Consistency is the reason why you should follow this rule.

README.md
volmer: Yes, and that's exactly why I'm changing the wording from a passive voice into the imperative tense. We shouldn't give space for doubt, the style guide should be as straightforward as possible.

README.md
volmer: I removed it because it's redundant with the rule below.

README.md
```diff
@@ -230,7 +228,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 ## Syntax
 
 * Use `::` only to reference constants (this includes classes and modules) and
-  constructors (like `Array()` or `Nokogiri::HTML()`). Do not use `::` for
+  constructors (like `Array()` or `Nokogiri::HTML()`). Avoid `::` for
```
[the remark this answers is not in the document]
volmer: It's not up to the style guide to make assumptions about "how strong this rule should be". I am using "avoid" as the consistent verb to advise developers about things they shouldn't do, and making exceptions depending on how strong we feel about a specific rule defeats the purpose of this very change.

rafaelfranca (approved):

README.md
PragTob: it _should_ extend to `nil` as `||` triggers that way for both `false` and `nil` so that the computation would be triggered again

README.md
PragTob: Like that line of thinking!

README.md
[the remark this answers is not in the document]
PragTob: Ah okay, then I misunderstood its purpose. In that case I think there are still some cases that are formulated in a way that suggests how strong this should be but this is good to ship without that imo

PragTob (approved):
WhatsApp Image 2020-09-18 at 07 53 37(2)

README.md
PragTob: I realize the intention of the PR was different and that's my misunderstanding, sorry - carry on :D

README.md
sambostock: The [difference involves precedence](https://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby), but it's not too hard to search for. Still, it might make sense to still have a link.

README.md
sambostock: Yeah. Consider

```ruby
def value
  @value ||= sleep(1000) && ['string', true, false, nil].shuffle
end
```

The truthy cases will be memoized, but the falsey ones won't be and would duplicate the computation, whereas this would memoize properly the first time:

```ruby
def value
  return @value if defined?(@value)
  @value = sleep(1000) && [true, false, nil].shuffle
end
```

Really, the rule should be

> Do not use `||=` to **memoize** potentially falsey variables (`false` or `nil`).
> &lt;code example demonstrating use of defined?&gt;

If the use case is potentially falsey memoization, then you **never** want to use `||=`. If you're using `||=` for something else, it **might** be fine.

README.md
sambostock: You might be right, as it can sometimes be a syntax error.

```ruby
def foo (a, b) # works
end

foo 1, 2       # works
foo(1, 2)      # works
foo (1, 2)     # SyntaxError: unexpected ',', expecting ')'
               # foo (1, 2)
               #       ^
```

volmer: From my previous reply:

> It's not up to the style guide to make assumptions about "how strong this rule should be". I am using "avoid" as the consistent verb to advise developers about things they shouldn't do, and making exceptions depending on how strong we feel about a specific rule defeats the purpose of this very change.

sambostock (approved):

PR #194, opened 2020-09-21
Merge "The Rest" into "General" section

volmer:
This change merges all rules from "The Rest" into "General", to make things a bit simpler. These are rules that do not fit necessarily in any of the other sections. If you believe any of these could be better placed in more specific sections please let me know. 

I am also omitting some rules from "General" that in my opinion are not related to this style guide. More on that in the comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * [Regular Expressions](#regular-expressions)
 * [Percent Literals](#percent-literals)
 * [Testing](#testing)
-* [The Rest](#the-rest)
 
 ## General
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1150,23 +1152,3 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
-
-## The Rest
-
-* Avoid long methods.
-
-* Avoid long parameter lists.
-
-* Avoid needless metaprogramming.
-
-* Avoid using `update_all`. If you do use it, use a scoped association
```
volmer, line 1162: Removing this rule since it is Rails specific.

README.md
```diff
@@ -73,15 +72,18 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Avoid monkeypatching.
 
-* Keep the code simple.
```
volmer, line 76: Removed the following rules since they are not about the "style in which you write Ruby" and more about general guidelines when programming. These are all very much implicit and I don't see the value of mentioning any of them:

* Keep the code simple.
* Do the right amount of design, not too little but also not too much.
* Avoid bugs.
* Be consistent.
* Use common sense.

sambostock (approved):

README.md
sambostock: Feels like there may be value in us creating a Rails style guide. Similarly to how we release this as the `rubocop-shopify` gem and depend on `rubocop`, we could have a `rubocop-rails-shopify` 🤔 

But yeah, I'm fine this style guide being strictly about Ruby.

etiennebarrie (approved):

PR #196, opened 2020-09-21
Update the CamelCase docs in README.md

cjhutchi:
Significant effort was done on the @Shopify/kernel-architecture-patterns team to remove custom inflections as they became problematic with the introduction of [Packwerk](https://www.github.com/Shopify/packwerk). 

Discussion has begun as to whether or not the Style Guide should be pushing acronyms to be in all caps (i.e. `HTTP`, `XML`, etc.) or follow the current state which doesn't require custom inflections (i.e. `Http`, `Xml`, etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,8 +441,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Use `snake_case` for symbols, methods, and variables.
 
-* Use `CamelCase` for classes and modules, but keep acronyms like HTTP, RFC, XML
-  uppercase.
+* Use `CamelCase` for classes and modules.
 
 * Use `snake_case` for naming files and directories, e.g. `hello_world.rb`.
 


rafaelfranca (approved):

cjhutchi:
@rafaelfranca @gmcgibbon brought up a valid point on Slack that the current Ruby standard involves capitalizing all acronyms. Figured we could take this discussion here for documentation purposes.

andyw8:
Should this only apply to Rails apps?

volmer:
Could you elaborate more on the rationale behind enforcing `Http` over `HTTP`? IMO acronyms should always be all caps for readability. Why would sacrifice readability in this case? Also would such concern be too Rails specific?

Keep in mind also that this style guide is public. The gem and team you referenced are in a private repository.

exterm:
I don't think defaulting to all-uppercase acronyms makes sense in code being autoloaded through zeitwerk, as each one needs its own custom inflection.

This increases the complexity of the code; if a developer wants to map a constant to a file name, they need to know whether a custom inflection is defined or not.

This may be a bit theoretical, but when statically analyzing the codebase it becomes much more obvious.

So I guess my point here is that all-uppercase acronyms are not a good default because they de facto are no longer the default in Rails applications. If you need to define an exception for each one, that can't be called a default.

gmcgibbon:
> Should this only apply to Rails apps?

This is my preference. We should say that camelcase is the standard for all _autoloaded_ code, and acronym case can be used for non-autoloaded lib code.

tomstuart:
I support this change, and personally feel that simple consistency (“always use `CamelCase`”) is more likely to be observed than more complex rules for autoloaded-vs-`require`d, Rails-vs-Ruby situations. I find `CamelCase` both less surprising, because it goes with the grain of Zeitwerk and therefore “just works” by default, and actually easier to read in practice, since e.g. `HTTPParser` creates a caps trainwreck that `HttpParser` avoids.

I think we should include a few words of explanation in the style guide, _at least_ calling out the acronym case in the form of illustrative examples, because otherwise it reads like a rule that is merely stating the obvious rather than taking a principled position. If there’s a concise way to mention autoloading conventions as a justification then that’d be even better.

andyw8:
If we go ahead, should we also add a custom RuboCop rule to enforce it?

volmer:
I like the argument about simplifying the current rule. In fact I agree we can omit the acronyms part of the rule, as it's being proposed in the PR right now. 

I wouldn't, however, enforce CamelCase for acronyms necessarily, because `HTTP` still reads better than `Http`, since it follows a standardized style we see in Ruby Core as well in the English language.

That being said, in the real world we have acronyms that don't follow such all-caps rule, such as `LotR`, `Unesco`, `Unicef`. There are also different styling recommendations in `en-US` and `en-GB`. Acronyms capitalization varies on a case-by-case basis, which complicates any effort of enforcing a single style in here. In the end, it would be up to the developer to decide how to write acronyms, according to each situation, and that's okay with me.

Also a reminder that consistency and readability are stronger arguments in this style guide than limitations in non-core Ruby tooling.

PR #206, opened 2020-10-20
Add link to shopify-cops gem

andyw8:
I stumbled across this recently, worth referencing here?


rafaelfranca:
Not yet. That gem is not recommended to be used on apps yet so I prefer if our style guide don’t promote it. This stance may change when we review what is our application style guide for the company that will include rails , performance and custom cops.

PR #231, opened 2021-02-08
Exclude node_modules directory

ChrisBr:
In a new Rails project I get these

```
node_modules/node-sass/src/libsass/extconf.rb:5:1: C: Style/GlobalVars: Do not introduce global variables.
$CFLAGS << " #{ENV['CFLAGS']}"
^^^^^^^
node_modules/node-sass/src/libsass/extconf.rb:6:1: C: Style/GlobalVars: Do not introduce global variables.
$LIBS << " #{ENV['LIBS']}"
^^^^^
```

Should we exclude the `node_directory` by default as it's a third party?

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -1,6 +1,7 @@
 AllCops:
   Exclude:
-  - 'db/schema.rb'
+    - 'db/schema.rb'
+    - 'node_modules/**/*'
   DisabledByDefault: true
   StyleGuideBaseURL: https://shopify.github.io/ruby-style-guide/
 


rafaelfranca:
I think this belongs to the .rubocop.yml config inside the app. We need to fix the app generator, not the styleguide gem.

andyw8:
RuboCop _should_ exclude `node_modules` by default so I wonder why it isn't in this case:

https://github.com/rubocop-hq/rubocop/blob/edef5e216f06f0848d60bb2961ec25bd94e0ee28/config/default.yml#L64

ChrisBr:
@andyw8 maybe because we need to enable inherit mode?

```
inherit_mode:
  merge:
    - Exclude
```
https://docs.rubocop.org/rubocop/configuration.html#merging-arrays-using-inherit_mode

PR #233, opened 2021-02-10
Add Layout/ArgumentAlignment

jbourassa:
Add the Layout/ArgumentAlignment rule definition, with params matching
~Layout/FirstArgumentIndentation~ Layout/ParameterAlignment. Prior to this
rule, this code would have been valid:

```ruby
method_call(
  1,
    2,
  3,
)
```

I argue that it should not be – params should all have the same
indentation.

By enforcing `with_fixed_indentation`, the following becomes invalid:
```ruby
puts 1,
     2
```

What do you think of this rule? If that's too strict, is there a way to support both
styles (`with_fixed_indentation` and `with_first_argument`), while still outlawing
the above offending code?

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -20,6 +20,12 @@ Style/Alias:
   - prefer_alias
   - prefer_alias_method
 
+Layout/ArgumentAlignment:
+  EnforcedStyle: with_fixed_indentation
+  SupportedStyles:
+    - with_first_argument
+    - with_fixed_indentation
+
 Layout/HashAlignment:
   EnforcedHashRocketStyle: key
   EnforcedColonStyle: key
@@ -164,14 +170,6 @@ Naming/FileName:
   ExpectMatchingDefinition: false
   Regex:
   IgnoreExecutableScripts: true
-
-Layout/FirstArgumentIndentation:
-  EnforcedStyle: consistent
-  SupportedStyles:
-  - consistent
-  - special_for_inner_method_call
-  - special_for_inner_method_call_in_parentheses
-  IndentationWidth:
 
 Style/For:
   EnforcedStyle: each


rafaelfranca:
Did you try to apply this new cop in Shopify/shopify? How many violations there were?

jbourassa:
>Did you try to apply this new cop in Shopify/shopify? How many violations there were?

I just did: 2268 offenses detected.

I made a spot check and noticed that some of the violations would be valid under `with_first_argument`, while some looked like random indentation and would've failed regardless of the chosen style.

volmer (commented):
I agree that the example you mentioned is problematic. That's because in regards to argument alignment we currently only have the [Layout/FirstArgumentIndentation](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/first_argument_indentation.rb) enabled with the `consistent` setting, but that cop only checks the first argument. 

I agree that we should enable [Layout/ArgumentAlignment](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/argument_alignment.rb) as well, so we enforce the wording we use in our Style Guide:

> Align the parameters of a method call, if they span more than one line, with one level of indentation relative to the start of the line with the method call.

However, notice that by setting it to `with_first_argument` you make the Layout/FirstArgumentIndentation cop obsolete, so we should disable that cop along with this change.

jbourassa:
>However, notice that by setting it to `with_first_argument` you make the Layout/FirstArgumentIndentation cop obsolete

Oh, I didn't know! Thanks for pointing it out – here's the [relevant part of the doc](https://github.com/rubocop-hq/rubocop/blob/5faffb3dc6df902feba92af469f1f1571bdd7d80/lib/rubocop/cop/layout/first_argument_indentation.rb#L13-L14):

```ruby
# This cop will respect `Layout/ArgumentAlignment` and will not work when
# `EnforcedStyle: with_fixed_indentation` is specified for `Layout/ArgumentAlignment`.
```

>so we should disable that cop along with this change.

✅ Done, PR updated.

rubocop.yml
```diff
@@ -165,12 +165,11 @@ Naming/FileName:
...
-  EnforcedStyle: consistent
+Layout/ArgumentAlignment:
+  EnforcedStyle: with_fixed_indentation
   SupportedStyles:
-  - consistent
-  - special_for_inner_method_call
-  - special_for_inner_method_call_in_parentheses
+    - with_first_argument
+    - with_fixed_indentation
   IndentationWidth:
```
volmer, line 173: Maybe we can remove this

```suggestion
```

rubocop.yml
```diff
@@ -165,12 +165,11 @@ Naming/FileName:
   Regex:
   IgnoreExecutableScripts: true
 
-Layout/FirstArgumentIndentation:
-  EnforcedStyle: consistent
+Layout/ArgumentAlignment:
```
volmer, line 168: Perhaps we can move this up to keep this config sorted alphabetically?

volmer (approved):
A few comments otherwise LGTM!

rafaelfranca (approved):

rubocop.yml
jbourassa: Quite a few rules have it empty, but [looks like empty (nil) is the same as not specifying](https://github.com/rubocop-hq/rubocop/blob/48c1e7a9e3b34ae145f4c20785a1bd04c7ec91d6/lib/rubocop/cop/mixin/alignment.rb#L14-L17). 👍

PR #234, opened 2021-02-10
Replace actions/setup-ruby with ruby/setup-ruby

sambostock:
[actions/setup-ruby is deprecated](https://github.com/actions/setup-ruby).

[ruby/setup-ruby](https://github.com/ruby/setup-ruby)...

- offers the [`bundler-cache: true`](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically) option to cache and install gems
  automatically, we don't need to do that manually anymore
- supports [inferring the Ruby version from `.ruby-version`](https://github.com/ruby/setup-ruby#supported-version-syntax)
- [recommends `@v1`](https://github.com/ruby/setup-ruby#versioning), instead of a more granular one, to get updates

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -9,18 +9,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Ruby 2.7
-      uses: actions/setup-ruby@v1.1.3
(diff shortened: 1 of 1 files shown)


volmer (approved):
Lovely! Thank you.

rafaelfranca (approved):

PR #235, opened 2021-02-10
Only build on `push`, not on `pull_request` too

sambostock:
Listening for both events triggers a redundant build. All PRs will have involved pushing a branch, so the build will have already been triggered. There is no need to trigger it again.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -1,6 +1,6 @@
 name: Ruby
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:


lugray:
This is a public repo. For external contributors, their pushes don't go to this repo, they go to their own clone, and so don't trigger CI. Once they open a PR on our repo, the PR triggers CI. For internal developers, builds on push are convenient to see failures before opening a PR.

dylanahsmith:
Here is how I accomplished this in the public identity_cache repo: https://github.com/Shopify/identity_cache/commit/4a55be5b81c4190d3aeb85534ed44977a05df912

PR #238, opened 2021-02-23
Enforce double quotes

miry:
There are often mix of quotes in same PR for same file.

Use single style for quotes in strings.
There is a ruby community gem `standard` for lint that suggests to use double quotes.

[Standard rule](https://github.com/testdouble/standard/blob/9b5f084390b4f41a1b679ab58721283c8ec86f1d/config/base.yml#L1124)
[Rails rule](https://github.com/rails/rails/blob/e2e34eaf8519f42d1d295688de27ea76280c276c/.rubocop.yml#L172)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-source 'https://rubygems.org'
+source "https://rubygems.org"
 
 gemspec
 
-gem 'github-pages', group: 'pages'
-gem 'rake'
+gem "github-pages", group: "pages"
+gem "rake"
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ We offer a default RuboCop configuration you can inherit from and be in sync
 with this Style Guide. To use it, you can add this to your `Gemfile`:
 
   ~~~ruby
-  gem 'rubocop-shopify', require: false
+  gem "rubocop-shopify", require: false
   ~~~
 
 And add to the top of your project's RuboCop configuration file:
@@ -146,24 +146,24 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   ~~~ ruby
   # starting point (line is too long)
   def send_mail(source)
-    Mailer.deliver(to: 'bob@example.com', from: 'us@example.com', subject: 'Important message', body: source.text)
+    Mailer.deliver(to: "bob@example.com", from: "us@example.com", subject: "Important message", body: source.text)
   end
 
   # bad (double indent)
   def send_mail(source)
     Mailer.deliver(
-        to: 'bob@example.com',
-        from: 'us@example.com',
-        subject: 'Important message',
+        to: "bob@example.com",
+        from: "us@example.com",
+        subject: "Important message",
         body: source.text)
   end
 
   # good
   def send_mail(source)
     Mailer.deliver(
-      to: 'bob@example.com',
-      from: 'us@example.com',
-      subject: 'Important message',
+      to: "bob@example.com",
+      from: "us@example.com",
+      subject: "Important message",
       body: source.text,
     )
   end
@@ -237,7 +237,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   ~~~ ruby
   # bad
   module A
-    FOO = 'test'
+    FOO = "test"
   end
 
   class A::B
@@ -246,7 +246,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
   # good
   module A
-    FOO = 'test'
+    FOO = "test"
 
     class B
       puts FOO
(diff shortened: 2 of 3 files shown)


lugray:
Single quotes convey useful information, namely, there is definitely no interpolation in this string. I often write PRs with a mix of single and double quotes. Single quotes whenever there's no interpolation, double quotes when I need them.

miry:
> Single quotes convey useful information, namely, there is definitely no interpolation in this string. I often write PRs with a mix of single and double quotes. Single quotes whenever there's no interpolation, double quotes when I need them.

Agree. I used the same approach before the `standard`. 

I found that new rails version 6.1 use double quotes as well.
That's why I sent PR to have this style.

I am ok with any approach.

Do you know what would be the best way to make decision in this style?
Is there some group that can choose which style is the best for Shopify?

lugray:
The style guide is useful for enforcing choices between multiple ways of doing things which are truly equivalent both in effect and in what they convey to a reader in order to make code more readable. That is not the case here. In this case, it's best left to developer discretion.  Enforcing the style I generally like to use is no better, as it would disallow:
```ruby
[
  "some string without interpolation",
  "some other string with #{interpolated}",
  "yet another string with #{interpolated}",
  "and many more strings with #{interpolated}",
]
```
Would that really be better with the first line single quoted? The lack of interpolation doesn't feel like it needs attention drawn to it. Conversely, in an array with only one value with interpolation, it seems useful to show its distinctness:
```ruby
[
  'an explicit string',
  'and another',
  'yet another',
  "whoops, this has #{interpolation} embedded in it",
  'you can see that in the left column',
]
```
This will always need a human decision maker. It's what we've been hired to do.

miry:
@lugray I found mostly in all pr that I review there are same simple stuff:

```ruby
a = 'string'

method "argument"
```

To format the code to have neither of styles with rubocop, the rule should be specified.
No matter of which style.

rafaelfranca:
I can see the value in the two camps, but recently I'm gravitating more towards double quotes everywhere.

The reason for that is that while I agree single quotes can convey useful information, the lack of consistency inside the same file and outside it makes the value of this information to be negated. We don't check if someone created a double quoted string with no interpolation so we end up with double quoted string that has no interpolation and the information conveyed by single quotes lose a little bit of its meaning.

lugray:
> the information conveyed by single quotes lose a little bit of its meaning.

I would say rather that the information conveyed by double quotes can loose meaning. Single quotes are always a guarantee that there's no interpolation.  And that seems useful.

rafaelfranca:
What I mean is that if double quotes can be also used for strings that have no interpolation, and people will gravitate towards consistency, "use single quotes if there is no interpolation" will not be followed, so you will get a lot of double quotes without interpolation and in the end you will have to get used to check the content of the string to make sure there is no interpolation.

masom:
For the sake of time saving on this and code uniformity, let's just go with double quotes and let rubocop auto-correct with `dev style` ?

jbourassa:
Opinion: I like the convenience of always having double quotes. No need to change the delimiter if you want to add an interpolation, nor escape `'`  in test names (when using the `test "something" do ...` definition style).

I personally value this simplicity more than the information carried by the author's choice of `"` or `'`.

volmer (commented):
After 6 years, it looks like we are ready to enforce a consistent quote style for Strings. For context, when we introduced this Style Guide in 2015 we decided to just tell devs to pick a style a be consistent with it: https://github.com/Shopify/ruby-style-guide/pull/7#discussion-diff-44665043

With the years we were able to successfully adopt this Style Guide across the organization, and we can introduce stricter rules sustainably thanks to gem versioning. Although my personal preference is for the single quote style, I gladly accept enforcing double quotes in name of consistency.

For this PR, updating the RuboCop config is not enough. We need to update the README with the new rule as well as add examples.

bdewater:
> Single quotes are always a guarantee that there's no interpolation. And that seems useful.

Seems or is useful? The only instance I can remember having to care for this was writing Rubocop rules (`on str` vs `on_dstr`), not during 'normal' development so I'm curious for any examples. My personal preference would be consistency and double quotes always 'just work'.

lugray:
> Seems or is useful? 

Is useful to me. `dev` by and large follows the convention I prefer, pretty consistently without a cop, and I find it useful.  I'm not interested in arguing it here anymore though. I'll just turn it off for `dev`.

rafaelfranca:
I think it is ok for existing apps to keep the current role if they wish for. I'd prefer if we applied this new role only for new apps since I'm sensing what people want is to not have to decide if they should enable this rule or not anymore.

We could add this role to our application generator by default, and that would satisfy this requirement while keeping exiting apps under "human discretion" like Lisa pointed out in a conversation with me.

But including this role in the style-guide allow us to document the way forward, so I'm more inclined in merge and release this change here and communicate to teams that it is ok, and even expected for them to make a call if they want to enforce this rule or not.

lugray:
While I still find myself in the human discretion camp, if there's going to be a rule, I think that the default configuration, preferring single quotes for non-interpolation strings is the more useful configuration.

volmer (approved):
Looks good! Thank you.

miry:
Let me know if I miss something.

rafaelfranca (approved):

PR #244, opened 2021-03-23
Allow RuboCop CLI config to inherit from the RuboCop Shopify gem

volmer:
The RuboCop CLI config was inheriting from the deprecated remote config via URL. This commit changes it to inherit from the RuboCop Shopify gem directly.


rafaelfranca (approved):

lugray:
I think this broke dev ci.  https://ruby-style-guide.shopify.dev/rubocop-cli.yml is now 404.

lugray:
...which seems to be a redirect from https://shopify.github.io/ruby-style-guide/rubocop-cli.yml

PR #245, opened 2021-03-23
Fix rubocop-cli.yml to inherit from rubocop.yml

macournoyer:
Current code causes an infinite recursion when inheriting from rubocop-cli.yml


volmer (approved):
Thank you! 

As a sidenote we could run RuboCop on CI with this config to make sure it works for next time.

PR #240, opened 2021-03-17
Rewrite website with Eleventy, Bulma, Prism, and friends

volmer:
Time to bring the Ruby Style Guide website to 2021.

This change replaces Jekyll with Eleventy, so we can benefit from NPM packages with Node. This allows us to use libraries such as Prism for syntax highlight and Bulma for CSS, among others.

With these libraries the website can be easily improved with new, better colours, more spacing and typography improvements for better readability.

Screenshots:

image

image

image

PR #246, opened 2021-03-23
Deploy rubocop.yml to Github pages

RouzbehMajidi:
I noticed after the latest deploy (I'm guessing https://github.com/Shopify/ruby-style-guide/pull/240) I could no longer access https://ruby-style-guide.shopify.dev/rubocop.yml
Screen Shot 2021-03-23 at 3 22 45 PM

I took a quick look and noticed the rubocop.yml file was missing from the `gh-pages` branch

I added `rubocop.yml` as a passthrough copy file in the eleventy config.

Tried it on my local (`npm run build && npm run serve`) and was able to access the rubocop.yml file

I also noticed the file `rubocop-cli.yml`, I've never used it but let me know if that should be deployed to and I can add it in

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -14,6 +14,7 @@ module.exports = function (eleventyConfig) {
   })
   eleventyConfig.setUseGitIgnore(false)
   eleventyConfig.addPassthroughCopy('CNAME')
+  eleventyConfig.addPassthroughCopy('rubocop.yml')
   eleventyConfig.addPassthroughCopy({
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request but that file not being available is by design. At this point, all the consumer of the style guide should be using the gem like explained here https://github.com/Shopify/ruby-style-guide#adoption-with-rubocop, not the yml file directly. Using the gem has as advantage to the application that changes to the style guide would not cause their CI to fail and they can adopt new rule as they upgrade the gem instead of when they are added.

PR #250, opened 2021-04-01
Inherit `Include` and `Exclude` paths from RuboCop

andyw8:
This was originally intended to prevent CI from inadvertently finding a `.rubocop.yml` in `node_modules` and applying that config.

The scope has expanded to provide a better out-of-the-box experience. When a user of the gem configures an `Include` or `Exclude`, this will now be merged with RuboCop's default configuration. Previously, this would have completely replaced RuboCop's defaults. 

This PR also removes the Rails-specific exclusion of `db/schema.rb` which should now be handled in the repo's own configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ And add to the top of your project's RuboCop configuration file:
     rubocop-shopify: rubocop.yml
   ~~~
 
+Any `Include` or `Exclude` configuration provided will be merged with RuboCop's defaults.
+
 For more information about inheriting configuration from a gem please check
 [RuboCop's
 documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-configuration-from-a-dependency-gem).
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -1,6 +1,9 @@
+inherit_mode:
+  merge:
+    - Exclude
+    - Include
+
 AllCops:
-  Exclude:
-  - 'db/schema.rb'
   DisabledByDefault: true
   StyleGuideBaseURL: https://shopify.github.io/ruby-style-guide/
 
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -59,7 +59,10 @@ AllCops:
   - "**/Vagabondfile"
   - "**/Vagrantfile"
   Exclude:
-  - "/db/schema.rb"
+  - "/node_modules/**/*"
+  - "/tmp/**/*"
+  - "/vendor/**/*"
+  - "/.git/**/*"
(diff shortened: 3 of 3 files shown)


andyw8:
Dupe of https://github.com/Shopify/ruby-style-guide/pull/231

rafaelfranca (approved):
Keeping the Rubocop defaults makes sense to me. I know those are application specific, but so it is db/schema.rb, so I think we would be better by using the same defaults as rubocop.

volmer (changes requested):
I 💯 agree that we should be excluding these directories, but that should be done by inheriting the default exclude files from RuboCop directly, just like we do currently with `Include` (which allows us to include files such as `Gemfile`, `Rakefile`, etc). 

We used to inherit `Exclude` in the past, until we created our own list in  #36. That's the point where we stopped inheriting from RuboCop as our list overrides theirs.

In order to add more patterns to `Exclude` as well as still inheriting the original list, we must use `inherit_mode`: https://docs.rubocop.org/rubocop/configuration.html#merging-arrays-using-inherit_mode

However, I am not sure if there is even a point in doing so. Note that all apps currently using our gem probably have their own `Exclude` lists, which completely overrides our own. That's the current behaviour. In other words, our exclusion of `db/schema.rb` is likely useless, as well as this new list in this PR.

Solving this is also not as simple as enabling `inherit_mode`, as this might arguably break the RuboCop config of current apps.

My opinion so far is that we should not define a `Exclude` config. RuboCop's defaults are good enough, and apps that want to use can simple set their own `inherit_mode`.

andyw8:
Thanks @volmer, good points there.

So although we may not be able to solve this for existing apps, could we improve the situation for new apps by changing the README instructions to:

```
inherit_mode:
  merge:
    - Exclude
inherit_gem:
  rubocop-shopify: rubocop.yml
```

volmer:
We could either give users instructions on inheritance mode or be opinionated about enabling the merging of those configs. But that opens up another question, which is: why would we enable (or suggest to enable) the merging of `Exclude` but not `Include` and other settings? RuboCop's default is the complete override for all settings in all cops when a custom config is defined. For sure nothing prevents us to consider `Include`/`Exclude` special cases; I am just pushing us to be intentional with our own decisions.

andyw8:
Update as per Slack with @volmer. Sharing for discussion, but I have not tested yet so should not be merged.

andyw8:
(I think an explanation should also be added to the README).

volmer (approved):
Looks good to me! 👍🏼

rafaelfranca (approved):

PR #253, opened 2021-04-06
Inherit supported styles config from RuboCop

volmer:
We are unnecessarily defining our own supported styles for cops. 

This was a leftover from the migration to `DisableByDefault` years back, when I dumped the entire config for all enabled cops. Although we do want to enable cops explicitly and set their preferred configs, we can still use RuboCop's supported styles config without breaking anything.

Thanks to the new test for config we can see that this change does not modify our resulting config in any significant way.

### DIFF
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -176,6 +176,7 @@ Layout/AccessModifierIndentation:
   SupportedStyles:
   - outdent
   - indent
+  IndentationWidth:
 Layout/ArgumentAlignment:
   Description: Align the arguments of a method call if they span more than one line.
   StyleGuide: "#no-double-indent"
@@ -204,6 +205,7 @@ Layout/AssignmentIndentation:
   Enabled: true
   VersionAdded: '0.49'
   VersionChanged: '0.77'
+  IndentationWidth:
 Layout/BeginEndAlignment:
   Description: Align ends corresponding to begins correctly.
   Enabled: false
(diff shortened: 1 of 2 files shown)


test/fixtures/full_config.yml
```diff
@@ -534,7 +543,7 @@ Layout/IndentationConsistency:
   EnforcedStyle: normal
   SupportedStyles:
   - normal
-  - rails
+  - indented_internal_methods
   Reference:
   - https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions
```
sambostock, line 548: Is this still relevant? Shouldn't we be pointing to our own style guide instead of the Rails one?

sambostock (commented):
I notice there are some **additions** on top of just deletions, as well as edits to sections other than `SupportedStyles`. I assume this is all just to get us back in line with defaults?

volmer:
> I notice there are some additions on top of just deletions, as well as edits to sections other than SupportedStyles. I assume this is all just to get us back in line with defaults?

Yes, those are all values that we can inherit directly from cops that do not necessarily impact what we are enabling/disabling or the rules we are enforcing.

test/fixtures/full_config.yml
volmer: This is unrelated with this PR, but for the reference https://github.com/rubocop/rubocop/pull/7059

rafaelfranca (approved):

PR #256, opened 2021-04-18
Add style word array bracket to .rubocop.yml

marcushwz:
I am not sure if I am missing anything, but according to the style guide, literal array syntax is preferred over `%i` and `%w`.
```
# bad
STATES = %w(draft open closed)

# good
STATES = ["draft", "open", "closed"]
```

- However, looking at the current Cop enabled, [Style/WordArray](https://msp-greg.github.io/rubocop/RuboCop/Cop/Style/WordArray.html) is not included inside.  I have extended it manually for my project, so that I will get linting message like this:
image

- Therefore, I am making this PR, to add in the `Style/WordArray` and set it to `bracket` as this is the preferred style as mentioned in the style guide.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -631,7 +631,7 @@ Style/SwapValues:
   Enabled: false
 
 Style/SymbolArray:
-  Enabled: false
+  EnforcedStyle: brackets
 
 Style/TrailingBodyOnMethodDefinition:
   Enabled: false
@@ -655,7 +655,7 @@ Style/UnpackFirst:
   Enabled: false
 
 Style/WordArray:
-  Enabled: false
+  EnforcedStyle: brackets
 
 Style/YodaCondition:
   Enabled: false
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -3401,10 +3401,10 @@ Style/SwapValues:
 Style/SymbolArray:
   Description: Use %i or %I for arrays of symbols.
   StyleGuide: "#percent-i"
-  Enabled: false
+  Enabled: true
   VersionAdded: '0.9'
   VersionChanged: '0.49'
-  EnforcedStyle: percent
+  EnforcedStyle: brackets
   MinSize: 2
   SupportedStyles:
   - percent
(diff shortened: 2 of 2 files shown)


rubocop.yml
```diff
@@ -1027,3 +1027,7 @@ Lint/DeprecatedOpenSSLConstant:
 
 Lint/Syntax:
   Enabled: true
+
+Style/WordArray:
+  EnforcedStyle: brackets
+  Enabled: true
```
volmer, line 1033: You don't need this line. By declaring the cop with a config it is going to be enabled already.

volmer (approved):
Thanks for catching this!

volmer:
@marcushwz Could you please resolve the conflicts and squash your commits? I'll merge it afterwards.

volmer:
Also could you please enable `SymbolArray` as well?

marcushwz:
@volmer Done adding in SymbolArray with brackets enforced style. I have two questions:
1. Do we still want the `Lint/Syntax` cop to be enabled? I noticed it was removed, I have added it back in but let me know if we don't want it to be included.
2. How do I update the `Description` in `full_config.yml`? I was thinking to update the Description to something like `Use [] for arrays of words`. But doing so will fail the `config test`. Then when I ran, `be rake config:dump` the description reverts back to original.

volmer:
> Do we still want the Lint/Syntax cop to be enabled? I noticed it was removed, I have added it back in but let me know if we don't want it to be included.

Yes, Lint/Syntax is now enabled by default. You don't need to enable it explicitly. You can notice that `full_config.yml` already has it enabled.

> How do I update the Description in full_config.yml? 

Don't worry about the description. By changing the config RuboCop will output the proper offence message accordingly.

marcushwz:
> > Do we still want the Lint/Syntax cop to be enabled? I noticed it was removed, I have added it back in but let me know if we don't want it to be included.
> 
> Yes, Lint/Syntax is now enabled by default (see #220). You don't need to enable it explicitly. You can notice that `full_config.yml` already has it enabled.
> 
> > How do I update the Description in full_config.yml?
> 
> Don't worry about the description. By changing the config RuboCop will output the proper offence message accordingly.

Okay, I have removed the `Lint/Syntax`, should be fine now. Let me know if you need me to change anything else. Thanks and have a great day!

volmer:
Thank you @marcushwz !

PR #273, opened 2021-05-31
Allow shadowing outer local variables

etiennebarrie:
This lint was created to replace a warning that Ruby removed.

The warning existed because there was a change in behavior introduced in version 1.9. Previously reusing a local variable resulted in the outer variable taking the value of the inner variable. After that change, there is shadowing, so they added a warning, which was finally removed in 2.6: https://bugs.ruby-lang.org/issues/12490.

But like mentioned in that issue, I think it's fair to shadow variables, like in the example:

```ruby
user = users.find { |user| user.parent == parent }
```

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -249,6 +249,9 @@ Lint/SendWithMixinArgument:
 Lint/ShadowedArgument:
   Enabled: false
 
+Lint/ShadowingOuterLocalVariable:
+  Enabled: false
+
 Lint/StructNewOverride:
   Enabled: false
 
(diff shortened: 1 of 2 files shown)


rafaelfranca:
I personally think that lint rule is still useful for who is reading the code. The interpreter knows what to do, but who is reading don't know what the interpreter will do.

etiennebarrie:
As far as I know most languages do it that way (shadowing does not impact the outer variable), so I don't think there's a lot of confusion to be had: https://en.wikipedia.org/wiki/Variable_shadowing.

Most importantly, there's nothing in the style guide about shadowing, so I don't think the rubocop rules should prevent it.

rafaelfranca (approved):
Yeah, it makes sense. This rule sometimes force us to write very confusing and forced code.

volmer (approved):

PR #288, opened 2021-06-30
Enable and configure Naming/InclusiveLanguage

tjwp:
This is a follow-up to https://github.com/Shopify/shopify-cops/issues/204.

A generic [Naming/InclusiveLanguage](https://github.com/rubocop/rubocop/pull/9842) cop has been included with the recent `v1.18.x` release of Rubocop. 

The cop can be configured to search in various places for problematic terms and suggest more inclusive language.

This pull request proposes that we enable the cop and explicitly set all configuration. I don't recommend relying on defaults for this cop because I expect they may continue to change in rubocop. A day after the initial release `master` was removed from the flagged terms, and `CheckStrings` was flipped to false.

I'm hoping that this pull request can be used to discuss the specifics of the configuration that we want to enable, especially with respect to the flagged terms.

Note: I'll update the config dump once there is agreement on the configuration.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -340,7 +340,17 @@ Naming/HeredocDelimiterNaming:
   Enabled: false
 
 Naming/InclusiveLanguage:
-  Enabled: false
+  Enabled: true
+  FlaggedTerms:
+    master:
+      Suggestions:
+        - main
+        - primary
+        - leader
+      AllowedRegex:
+        - !ruby/regexp '/master[_\s\.]key/' # Rails master key
+        - 'blob/master/'
+        - 'origin/master'
 
 Naming/MemoizedInstanceVariableName:
   Enabled: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (2.1.0)
-      rubocop (~> 1.17)
+      rubocop (~> 1.18)
 
 GEM
   remote: https://rubygems.org/
--- a/rubocop-shopify.gemspec
+++ b/rubocop-shopify.gemspec
@@ -21,5 +21,5 @@ Gem::Specification.new do |s|
     "allowed_push_host" => "https://rubygems.org",
   }
 
-  s.add_dependency("rubocop", "~> 1.17")
+  s.add_dependency("rubocop", "~> 1.18")
 end
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -1857,7 +1857,7 @@ Naming/HeredocDelimiterNaming:
   - !ruby/regexp /(^|\s)(EO[A-Z]{1}|END)(\s|$)/
 Naming/InclusiveLanguage:
   Description: Recommend the use of inclusive language instead of problematic terms.
-  Enabled: false
+  Enabled: true
   VersionAdded: '1.18'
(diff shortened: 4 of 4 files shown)


rubocop.yml
```diff
@@ -336,6 +336,41 @@ Naming/HeredocDelimiterCase:
...
+Naming/InclusiveLanguage:
+  Enabled: true
+  CheckIdentifiers: true
+  CheckConstants: true
+  CheckVariables: true
+  # CheckStrings is false in the rubocop defaults
+  CheckStrings: true
+  CheckSymbols: true
+  CheckComments: true
+  CheckFilepaths: true
```
volmer, line 348: IMO this new cop (which helps us enforce the rule introduced by #167, by the way) makes sense for naming in variables, constants, etc, but when it comes to the contents of strings it doesn't make much sense to me. I think we should omit this configs and stick with RuboCop's defaults.

rubocop.yml
```diff
@@ -336,6 +336,41 @@ Naming/HeredocDelimiterCase:
...
+        - leader
+      AllowedRegex:
+        - !ruby/regexp '/master[_\s\.]key/' # Rails master key
+        - 'blob/master/'
+        - 'origin/master'
+    slave:
+      Suggestions:
+        - replica
+        - secondary
+        - follower
```
volmer, line 372: Why aren't these the defaults from RuboCop itself?

rubocop.yml
tjwp: My counterargument for including strings is that it catches references in error messages ("add xyz to whitelist to allow ...") and test descriptions. Test descriptions often have references to the same identifiers, constants, variable where we'd look for offenses.

rubocop.yml
volmer: Got it. I think we should only define the configs we are overriding. If you dump the full config we can assert that the merged result is the expected one during the review (and also make the PR pass CI).

rubocop.yml
```diff
@@ -336,6 +336,21 @@ Naming/HeredocDelimiterCase:
...
   Enabled: false
 
+Naming/InclusiveLanguage:
+  Enabled: true
+  FlaggedTerms:
+    whitelist:
+      AllowedRegex:
+        - 'web_console.whitelisted_ips'
+    master:
+      Suggestions:
```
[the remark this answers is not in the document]
tjwp: Added. Context definitely matters here but that is difficult for a cop to determine. `main` makes sense for a branch, but `primary` or `leader` make sense for a node or server.

rubocop.yml
```diff
@@ -336,6 +336,21 @@ Naming/HeredocDelimiterCase:
 Naming/HeredocDelimiterNaming:
   Enabled: false
 
+Naming/InclusiveLanguage:
+  Enabled: true
+  FlaggedTerms:
+    whitelist:
+      AllowedRegex:
+        - 'web_console.whitelisted_ips'
```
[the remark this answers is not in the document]
tjwp: Happy to remove it, but this was included in Shopify Rails apps generated prior to March: https://github.com/Shopify/shopify-rails/pull/399

The configuration was deprecated in web-console in v4.0.3, but is currently still supported.

volmer (approved):
Looks great, thank you for the PR and thanks for creating the cop. Please don't forget to squash your commits 🙏🏼

rafaelfranca (approved):

PR #298, opened 2021-07-26
Enable Style/Documentation

tjwp:
If it would be useful for every class, and non-namespace module, to have a comment, then enabling this cop would help to encourage that behavior and signal to developers that these kinds of comments are wanted. 

CC @exterm

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -431,7 +431,7 @@ Style/DocumentDynamicEvalDefinition:
   Enabled: false
 
 Style/Documentation:
-  Enabled: false
+  Enabled: true
 
 Style/DoubleCopDisableDirective:
   Enabled: false
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -2357,7 +2357,7 @@ Style/DocumentDynamicEvalDefinition:
   VersionChanged: '1.3'
 Style/Documentation:
   Description: Document classes and non-namespace modules.
-  Enabled: false
+  Enabled: true
   VersionAdded: '0.9'
   AllowedConstants: []
   Exclude:


tjwp:
> You mention
> 
> > every non-namespace module
> 
> is this something Rubocop supports out of the box?

Yes, I was paraphrasing from their documentation: https://docs.rubocop.org/rubocop/cops_style.html#styledocumentation
Namespace modules (defined below) are excluded:

> This cop checks for missing top-level documentation of classes and modules. Classes with no body are exempt from the check and so are namespace modules - modules that have nothing in their bodies except classes, other modules, constant definitions or constant visibility declarations.

exterm:
Going to copy over my comment from slack

> I am convinced that people will only write good, helpful, non-brittle comments if they understand why they are important. That’s why I posted this; I thought it’d make a good example. Still, even if they know why they are important, they might forget, so a trigger can be helpful.
However, the trigger without the understanding can have a negative effect.
This is actually really well described by B.J. Fogg’s behavioral model that I also used in my blog post last year https://shopify.engineering/shopify-monolith

Fogg-Behavior-Model

In conclusion, enabling this cop in isolation is likely to cause problems (proliferation of low quality comments). I agree that we should at least provide some good examples, but I guess we can't easily link them from the error message?

🤔 Overall I am skeptical that just enabling the cop will work, but I'm OK with trying it out if someone can commit to following up after a few weeks and removing the cop if necessary.

The root problem IMO is one that can only be solved through teaching and communication. We need people to set good examples in production code, PR reviews and pairing sessions.

tjwp:
The message from this cop does not provide much guidance:

`'Missing top-level %<type>s documentation comment.'`

I agree that this needs to be paired with more information about what makes a good top-level comment. If the cop could be configured to provide a custom message that would make it more useful. But there is still limited information that can be conveyed through an offense message.

volmer (changes requested):
Documentation is not part of this Style Guide. The RuboCop config should help us enforce the rules we have in the Ruby Style Guide. 

Personally I always enable this cop because I find documentation highly valuable for my teams' projects, so I would gladly welcome documentation to be encourage across Shopify. But in order to enable this cop, we must discuss changing the Style Guide first with proper guidelines and explanation.

exterm:
Trying to demonstrate what I have in mind for module level comments in https://github.com/Shopify/shopify/pull/305142 and https://github.com/Shopify/shopify/pull/305147

tjwp:
> Documentation is not part of this Style Guide. The RuboCop config should help us enforce the rules we have in the Ruby Style Guide.
> 
> But in order to enable this cop, we must discuss changing the Style Guide first with proper guidelines and explanation.

There is a Comments section in the Style Guide so I have included a commit to add an item there for the suggested class/module comments.

@volmer if you'd prefer this as a separate PR for discussion on the Style Guide, I can do that. (Or the cop configuration could be dropped in favor of the added guidance.)

And for anyone reading this, please wordsmith me if you have suggestions or better examples. I was trying to keep things general, but this may be too contrived.

README.md
```diff
@@ -520,6 +520,23 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Avoid superfluous comments. Focus on **why** the code is the way it is if
   this is not obvious, not **how** the code works.
 
+* Short, top-level comments on classes and modules (that are more than a namespace) are
+  encouraged. Again, these comments should be about the purpose of the relevant code,
+  not a description of the implementation which will be brittle to refactorings and grow
+  stale quickly. Good comments on design intent support the long-term understanding
+  and maintainability of code.
```
exterm, line 527: ❤️

PR #303, opened 2021-08-02
Allow Trailing Whitespace in HEREDOC

dougedey-shopify:
When implementing this rule in Shopify/Shopify I found that it breaks a lot of assumptions with HEREDOC and signficantly reduces the readability of code.

This is the current `AllowInHeredoc: false` PR: https://github.com/Shopify/shopify/pull/305742/

You can see that it adds `#{"      "}` type changes in heredoc, but also auto-corrects the HEREDOC to be further indented (I'm not sure _why_) and that breaks a lot of test assumptions.

Here is the same PR but with `AllowInHeredoc: true`: https://github.com/Shopify/shopify/pull/305743

You can see this is fixing what I would expect 99%+ of developers would want fixed. And doesn't break tests/behaviour

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -108,6 +108,10 @@ Layout/SpaceBeforeBrackets:
 Layout/SpaceInLambdaLiteral:
   Enabled: false
 
+Layout/TrailingWhitespace:
+  Enabled: true
+  AllowInHeredoc: true
+
 Lint/AmbiguousAssignment:
   Enabled: false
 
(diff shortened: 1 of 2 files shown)


volmer (changes requested):
Hey! I think the current config for the trailing whitespace is fine, even for heredocs.

I understand that it might seem weird at a first glance, but autocorrect is blindly replacing the current trailing whitespaces with `#{' '}` in order to preserve the existing formatting for the strings in which heredoc is being used. Note however that in order to fix those instances in Shopify Core you should simply remove the trailing whitespaces manually, since for those specific cases the trailing whitespaces have no importance in formatting whatsoever as far as I could check. Autocorrect tries its best to fix offenses but it is ultimately the developer's responsibility to properly apply the rule while keeping the code sane.

Also note that this is a public repository, so not all contributors might have access to the PRs that you referenced.

PR #314, opened 2021-09-06
Enable `Style/ExplicitBlockArgument` cop

casperisfine:
The code is disabled by default because in some cases it could change the block arity which might fail if the block is
actually a lambda.

However I enabled it on [Rails](https://github.com/rails/rails/pull/43170) and [shopify-core](https://github.com/Shopify/shopify/pull/310915) and it went just fine.

Since this cop is beneficial for production performance and for trimming CI backtraces a bit, I think we should enable it.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -468,7 +468,7 @@ Style/ExpandPathArguments:
   Enabled: false
 
 Style/ExplicitBlockArgument:
-  Enabled: false
+  Enabled: true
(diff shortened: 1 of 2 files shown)


sambostock (approved):

rafaelfranca (approved):
Sorbet is another reason to have the explicit block argument. Without it, we can write the method signatures.

PR #316, opened 2021-09-16
Enable Bundler cops

volmer:
This is the first of a series of proposals intended to revisit the cops we are disabling for each department and see if they should be enforced. The idea is to see the value some disabled cops might bring to our users compared to the effort required to apply corrections to those.

This PR enables four cops from the `Bundler` department:

* There should be no duplicated gems in your `Gemfile`.
* Your Gemfile should be named `Gemfile` as well as your lock file should be named `Gemfile.lock`.
* Sources should be explicit with `https`.
* Gems should be listed in alphabetical order in your `Gemfile`.

More information about these cops can be seen at https://docs.rubocop.org/rubocop/cops_bundler.html

If this proposal is approved these cops will be enabled in the next minor version of `rubocop-shopify`.


test/fixtures/full_config.yml
```diff
@@ -138,15 +138,15 @@ Bundler/InsecureProtocolSource:
...
-  Enabled: false
+  Enabled: true
   VersionAdded: '0.50'
   Include:
   - "**/*.gemfile"
   - "**/Gemfile"
   - "**/gems.rb"
 Bundler/OrderedGems:
   Description: Gems within groups in the Gemfile should be alphabetically sorted.
-  Enabled: false
```
sambostock, line 149: While many projects have somewhat disorganized Gemfiles, many projects have tried to organize their Gemfiles into sections by what type of Gem it is (e.g. grouping gems that pertain to datastores).

While this rule could be disabled on a per-repo basis, is our "new" stance that we prefer organizing alphabetically over by category?

test/fixtures/full_config.yml
```diff
@@ -110,7 +110,7 @@ Bundler/GemComment:
   OnlyFor: []
 Bundler/GemFilename:
   Description: Enforces the filename for managing gems.
-  Enabled: false
```
sambostock, line 113: Presumably this rule would be disabled in projects using multiple Gemfiles for testing purposes?

test/fixtures/full_config.yml
tjwp: My experience with this cop is that it will only require gems to be sorted within "sections", so if those sections are separated by a blank line (or even a comment based on the default configuration), it will only require ordering within those sections. "group" in the description of the cop is generic and does not refer to the Gemfile `group` method.

We may want to consider changing the configuration for `TreatCommentsAsGroupSeparators` to false.

With the default true setting we can get a strange ordering:
```ruby
gem "a"
gem "s"
gem "x"
# comment on "b"
gem "b"
gem "t"
gem "y"
```

If it is set to false, then the auto-correct is smart enough to move comments with the gem that they precede.

rafaelfranca:
> * Gems should be listed in alphabetical order in your `Gemfile`.

I don't think this should be anabled. The order gems are defined in the Gemfile also dictates which order they are loaded in the process. There are a lot of subtile problems if for example you load Rails after other gems.

volmer:
Fair enough, keeping `Bundler/OrderedGems` disabled.

tjwp (approved):

PR #319, opened 2021-09-30
Enable Layout cops

volmer:
This is the third of a series of proposals intended to revisit the cops we are disabling for each department and see if they should be enforced. The idea is to see the value some disabled cops might bring to our users compared to the effort required to apply corrections to those.

This PR enables four cops from the `Layout` department, and reflects these changes in the Style Guide under the "Layout
 section (formerly known as the "Formatting" section) for the cops that require clarification or examples. Even though this list is a bit long, these cops are autocorrectable and easy to adopt. Some of these are also just common sense like avoiding leading spaces in source files or empty comment lines.

* Heredocs must be indented (added example to the Style Guide).
* Empty comment lines should be avoided.
* There should be an empty line between magic comments and the rest of the code (added example to the Style Guide).
* Empty lines around arguments should be avoided.
* Attribute accessors should be surrounded by empty lines (added example to the Style Guide).
* There should not be empty lines around `begin` and exception capture keywords (added example to the Style Guide).
* The first parameter of a method call should be aligned according to the previous line (the rule already exists).
* Leading empty lines should be removed.
* String concatenation should be indented when spanning multiple lines.
* There shouldn't be spaces around the method call operator (added example to the Style Guide).
* There shouldn't be spaces in lambda literals (added example to the Style Guide).

More information about these cops can be seen at https://docs.rubocop.org/rubocop/cops_layout.html

If this proposal is approved these cops will be enabled in the next minor version of `rubocop-shopify`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 ## Table of Contents
 
 * [General](#general)
-* [Formatting](#formatting)
+* [Layout](#layout)
 * [Syntax](#syntax)
 * [Naming](#naming)
 * [Classes and Modules](#classes-and-modules)
@@ -88,7 +88,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Avoid more than three levels of block nesting.
 
-## Formatting
+## Layout
 
 * Use `UTF-8` as the source file encoding.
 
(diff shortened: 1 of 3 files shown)


andyw8:
I'm sure you've already tested this against Core but it would interesting to see a branch showing what changes the auto-correct would make with these cops enabled.

sambostock (approved):

PR #332, opened 2021-11-08
Regex improvements

bdewater:
Regex anchoring was missing an IMO pretty important _why_ this is preferred, so added that.

For the `match?` addition I considered merging it with the non-capturing group recommendation, but figured both recommendations could stand on their own: one could be using both capturing and non-capturing groups in the same regular expression.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1199,7 +1199,8 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   string["text"]
   ~~~
 
-* Use non-capturing groups when you don't use the captured result.
+* Use non-capturing groups when you don't use the captured result inside the
+  regular expression.
 
   ~~~ ruby
   # bad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1199,7 +1199,8 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   string["text"]
   ~~~
 
-* Use non-capturing groups when you don't use the captured result.
+* Use non-capturing groups when you don't use the captured result inside the
```
sambostock, line 1202: Can you explain this change?

It seems to me like you could still want capturing groups even if you don't refer to them in the regular expression (i.e. `\1`, `\2`, etc.).

```ruby
def extract_date(string = "2021-11-08")
  match = /(\d\d\d\d)-(\d\d)-(\d\d)/.match(string)
  raise "invalid date" if match.nil?
  year, month, day = match.captures
  validate_date_pieces!(year, month, day)
  match.string
end
```

README.md
```diff
@@ -1209,6 +1210,16 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
   /(?:first|second)/
   ~~~
 
+* Use `Regexp#match?` when you don't use the captured result in Ruby.
```
sambostock, line 1213: Do you think it's worth mentioning why?

```suggestion
* Use `Regexp#match?` when you don't use the captured result in Ruby, to avoid creating the more expensive `MatchData` object.
```

README.md
volmer, line 1202: Same as @sambostock, didn't really get the reason behind this addition.

volmer (approved):
One small nitpick but overall I like the additions 👍🏼

PR #346, opened 2022-01-12
Ignore OpenStructUse for tests folders

miry:
Recently found there are a lot of project with OpenStruct in tests.
Propose to postpone update tests to remove blockers for developers
to rewrite a lot of tests.
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/OpenStructUse


miry:
I am not fun of usage Openstruct. Suggestion of cop to replace with `double`, that has the same problems as explained before. My point the alternative is not better. As I understand the main point of Struct vs Openstruct is performance.

Issue #189, opened 2020-09-15
Add more good vs bad examples


volmer:
Extracted from #74. It would be great to know exactly which rules would be made easier to understand with more examples. Some are simple and straightforward enough and don’t need examples.

cc @sambostock

PR #347, opened 2022-01-12
Introduce example to run validation of of rubocop rules

miry:
There are sometime misconfigured Readme and Rubocop.
Make sure Rubocop correctly work, propose to have a file with
good and bad cases.

Started work, but would like to get feedback, before migrate Readme examples to ruby file.

Partialy covers https://github.com/Shopify/ruby-style-guide/issues/189

### DIFF
--- a/examples/sample.rb
+++ b/examples/sample.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# rubocop:disable Lint/UselessAssignment
+
+# Prefer the literal array syntax over `%w` or `%i`.
+word_array_percent = ["red", "blue", "white"] # vs %w[red blue white]
+
+# Use literal array and hash creation notation unless you need to pass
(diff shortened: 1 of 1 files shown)


sambostock:
I wonder if it would be possible to inline these in the README itself and then run Rubocop against that. Maybe we could look for ```` ```ruby```` code blocks that have `# good` or `# bad` at the top.

Moving them to a seperate file just increases the synchronization problem, as we'd now have to keep three files in sync ( `README.md`, `examples/sample.rb`, and `rubocop.yml`) instead of just two.

volmer:
It would be really nice to be able to run RuboCop against the examples we have in the Readme. That would be better than having a separate Ruby file with code samples. Would you be interested in working on a parser for that?

miry:
It would be interesting.

PR #348, opened 2022-01-12
Enable `Layout/EmptyLineAfterGuardClause` cop

sambostock:
This enables the [`Layout/EmptyLineAfterGuardClause`](https://docs.rubocop.org/rubocop/cops_layout.html#layoutemptylineafterguardclause) cop, which is **autocorrectable**.

From the documentation

> ```ruby
> # bad
> def foo
>   return if need_return?
>   bar
> end
> 
> # good
> def foo
>   return if need_return?
> 
>   bar
> end
> 
> # good
> def foo
>   return if something?
>   return if something_different?
> 
>   bar
> end
> 
> # also good
> def foo
>   if something?
>     do_something
>     return if need_return?
>   end
> end
> ```

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -25,7 +25,7 @@ Layout/CaseIndentation:
   EnforcedStyle: end
 
 Layout/EmptyLineAfterGuardClause:
-  Enabled: false
+  Enabled: true
 
 Layout/EndAlignment:
   EnforcedStyleAlignWith: variable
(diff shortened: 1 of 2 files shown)


rafaelfranca:
How many violations for this cop do we have in core?

sambostock:
@rafaelfranca, [8,271 offenses detected](https://buildkite.com/shopify/shopify-linters/builds/200155), all auto-correctable though

nightsurgex2 (approved):

PR #365, opened 2022-03-02
Disable EnforcedShorthandSyntax in Style/HashSyntax

rafaelfranca:
We should not be forcing the shorthand syntax because of a few reasons:

1. That syntax doesn't always yields the best/clear code. It has
surprising behavior specially with method calls using the key names.
2. Ruby 3.1 shorthand syntax is still not supported by Sorbet.

We can fix 2, but not 1, so it is better to just allow people to chose
when to use the shorthand syntax in the context that makes sense.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -484,6 +484,9 @@ Style/HashExcept:
 Style/HashLikeCase:
   Enabled: false
 
+Style/HashSyntax:
+  EnforcedShorthandSyntax: either
+
 Style/HashTransformKeys:
   Enabled: false
 
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -2718,7 +2718,7 @@ Style/HashSyntax:
   - hash_rockets
   - no_mixed_keys
   - ruby19_no_mixed_keys
-  EnforcedShorthandSyntax: always
+  EnforcedShorthandSyntax: either
   SupportedShorthandSyntax:
   - always
   - never


rafaelfranca:
ruby19 seems better to me. If keys needs to be string, they still need to use hash rocket, so we sometimes need to mix syntax. Although that only happens if people are creating hashes that hold two different key types, which is so rare that I don't think we should care about that case.

volmer (approved):

tomstuart:
@rafaelfranca core currently has `EnforcedShorthandSyntax: never` in its RuboCop configuration, presumably to avoid the `always` default causing violations. If we [upgrade `rubocop-shopify` to v2.5.0 in core](https://github.com/Shopify/shopify/compare/upgrade-rubocop-shopify), can we remove that configuration entirely (i.e. inherit `either`) or would you like to keep it pinned to the more conservative `never`?

sambostock:
I suspect we'll still leave it up to the author to choose.

One thing I've noticed is that some issues & PRs cite this PR saying that "shorthand syntax should not be used", which as far as I can tell is incorrect. If I'm not mistaken, our stance is that we don't **enforce** shorthand syntax, but we do **allow** it, at the authors discretion, with the expectation that they go with what is clearest in that context.

```ruby
x, y, z = string.split(",").map(&:to_f)
Point.new(x: x, y: y, z: z) # fine
Point.new(x:, y:, z:)       # also fine
```

PR #366, opened 2022-03-03
Release version 2.5.0

tomstuart:
We need to bump the version number to get Shipit to publish the gem. The [tag](https://github.com/Shopify/ruby-style-guide/tree/v2.5.0) and [release](https://github.com/Shopify/ruby-style-guide/releases/tag/v2.5.0) already exist, and therefore won’t point to this commit, but I don’t think that will cause any real problems.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (2.5.0)
-      rubocop (~> 1.24)
+      rubocop (~> 1.25)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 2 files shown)


sambostock:
Can we/should we move the tag? Seems weird that the 2.5.0 tag would point to code with a different version.

tomstuart:
> Can we/should we move the tag after merging? Seems weird that the 2.5.0 tag would point to code with a different version.

We can, and (in my opinion) we should, but I don’t know if that’ll cause problems anywhere. I’ll wait for @rafaelfranca’s review.

paracycle:
Btw, you don't need to manually tag repos when doing a release, Shipit creates and pushed a tag when it is deploying the gem to Rubygems.

rafaelfranca:
It doesn't sign the tag though, so I always create manually to make sure the tag was signed.

PR #373, opened 2022-04-01
Remove redundant configuration keys

volmer:
There are some cops we are explicitly enabling/disabling although they are already enabled/disabled by default, as proven by our full config test. I am adding also another test to fail in case redundant entries are introduced in our config file.

### DIFF
--- a/test/config_test.rb
+++ b/test/config_test.rb
@@ -30,4 +30,29 @@ class ConfigTest < Minitest::Test
       assert(diff.empty?, error_message)
     end
   end
+
+  def test_config_has_no_redundant_entries
+    config = RuboCop::ConfigLoader.load_file("rubocop.yml")
+    default_config = RuboCop::ConfigLoader.default_configuration
+
+    # This entry is not a cop.
+    config.delete("inherit_mode")
+
+    config.each do |cop_name, cop_config|
+      default_cop_config = default_config.fetch(cop_name)
+      cop_config.each do |key, value|
+        error_message = <<~ERROR
+          Error: #{cop_name} was configured with the same value as the default
+          RuboCop configuration.
+
+          #{cop_name}:
+            #{key}: #{value}
+
+          Please remove the configuration as it is unnecessary.
+        ERROR
+
+        refute_equal(default_cop_config[key], value, error_message)
+      end
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


rafaelfranca (approved):

test/config_test.rb
```diff
@@ -30,4 +30,29 @@ def test_config_is_unchanged
...
+
+          #{cop_name}:
+            #{key}: #{value}
+
+          Please remove the configuration as it is unnecessary.
+        ERROR
+
+        refute_equal(default_cop_config[key], value, error_message)
+      end
+    end
```
sambostock, line 56: What do you think of the following?

```suggestion
  def test_config_has_no_redundant_entries
    config = RuboCop::ConfigLoader.load_file("rubocop.yml")
    default_config = RuboCop::ConfigLoader.default_configuration
    redundant_config = Hash.new { |hash, key| hash[key] = {} }

    # This entry is not a cop.
    config.delete("inherit_mode")

    config.each do |cop_name, cop_config|
      default_cop_config = default_config.fetch(cop_name)
      cop_config.each do |key, value|
        default_value = default_cop_config[key]
        redundant_config[cop_name][key] = value if value == default_value
      end
    end

    flunk(<<~ERROR.chomp) unless redundant_config.empty?
      Error: The following config values are the same as the default Rubocop configuration:

      #{redundant_config.to_yaml.sub(/\A---\n/, "").chomp}

      Please remove these entries from the configuration, as they are unnecessary.
    ERROR
  end
```

The difference being that this implementation collects all the redundant config and presents it at once, instead of failing after only the first entry.

#### Current implementation's failure message

```
Error: Layout/EmptyLineAfterGuardClause was configured with the same value as the default
RuboCop configuration.

Layout/EmptyLineAfterGuardClause:
  Enabled: true

Please remove the configuration as it is unnecessary.
.
Expected true to not be equal to true.
```

#### Proposed implementation's failure message

```
Error: The following config values are the same as the default Rubocop configuration:

Layout/EmptyLineAfterGuardClause:
  Enabled: true
Style/AsciiComments:
  Enabled: false
Style/ExplicitBlockArgument:
  Enabled: true
Style/GlobalStdStream:
  Enabled: true

Please remove these entries from the configuration, as they are unnecessary.
```

It also uses `flunk` to fail the test, avoiding the redundant `Expected X to not be X.` message.

volmer: I like it, thank you @sambostock ! ❤️

sambostock (approved):

PR #377, opened 2022-04-22
Rename deprecated `IgnoredPatterns` config to `AllowedPatterns`

### LINKED ISSUE #376: Layout/LineLength obsolete parameter `IgnoredPatterns`
Just noticed there's a warning showing up while running `rubocop` due to a deprecated(?) `IgnoredPatterns` param.

```
❯ rubocop
Warning: obsolete parameter `IgnoredPatterns` (for `Layout/LineLength`) found in /Users/larouxn/.gem/ruby/3.1.2/gems/rubocop-shopify-2.5.0/rubocop.yml
`IgnoredPatterns` has been renamed to `AllowedPatterns`.
Warning: obsolete parameter `IgnoredPatterns` (for `Layout/LineLength`) found in .rubocop.yml
`IgnoredPatterns` has been renamed to `AllowedPatterns`.
Inspecting 2166 files
```

https://github.com/Shopify/ruby-style-guide/blob/677c98edbe8bfa16c2536ffb90607233f07293a7/rubocop.yml#L46-L49

sambostock:
Also bumps Rubocop version.

Closes #376
Caused by rubocop/rubocop#10555
Closes #380 

We _could_ also switch to the `!ruby/regexp` syntax for the config, instead of a `String`, which would remove the need for all the escaping and clarify what the pattern is doing, but I can do that in a different PR.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (2.5.0)
-      rubocop (~> 1.25)
+      rubocop (~> 1.28)
 
 GEM
   remote: https://rubygems.org/
@@ -14,7 +14,7 @@ GEM
     method_source (1.0.0)
     minitest (5.15.0)
     parallel (1.22.1)
-    parser (3.1.1.0)
+    parser (3.1.2.0)
       ast (~> 2.4.1)
     pry (0.13.1)
       coderay (~> 1.1)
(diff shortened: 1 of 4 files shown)


sambostock:
I think this charge will require bumping the minimum Rubocop version (unless we can configure dynamically based on Rubocop version), and update the test that checks the full config.

sambostock:
🏓 @rafaelfranca & @volmer

sambostock:
#380 takes an alternative approach by using ERB to decide between configs by checking the Rubocop version, which allows us to handle the deprecation without forcing consumers to bump their Rubocop version.

However, I'm wondering if that matters at all? 🤔 Are we fine with requiring a version bump?

paracycle (approved):
I think a dependency version bump makes sense.

adrianna-chang-shopify (approved):

gjtorikian:
Could a version bump and release happen? 😅

PR #387, opened 2022-05-11
Enforce Multiline Method calls, Arrays, and Hashes to use one line per element

Korri:
This might be a personal pet peeve of mine, but I find all the examples inside the `# bad` to be hard to read, and even lead me to simply not realize that one of the elements is here at all on more complex cases.

Adding the rules on this PR will autocorrect all those cases to use one line per argument and/or element, leading to a much better readability!

Those [are actually recommended](https://docs.rubocop.org/rubocop/cops_layout.html#layoutlinelength) when using `Layout/LineLength`, and it does seems to be where I see the most of cases of "autocorrect gone wrong".

### bad (good before this Change)
```ruby
foo(
a,
b,
c,
)

foo(a, b,
  c)

foo(a,
  b,
  c)

{
  a: 1, b: 2,
  c: 3,
}

{ a: 1,
  b: 2,
  c: 3, }

[
  a, b,
  c,
]

[a,
  b,
  c,]
```
### Good (and enforced now)
```ruby
foo(
  a,
  b,
  c
)

foo(
  a,
  b,
  c
)

foo(a, b, c)

{
  a: 1,
  b: 2,
  c: 3,
}

{ a: 1, b: 2, c: 3 }

[
  a,
  b,
  c,
]

[a, b, c]
```

### Good (Exceptions, not affected by changes)
```ruby
# Multiline last argument
after_commit :do_something, if: -> {
  some_complex_logic
}

# Single argument call
foo({
  bar: bat,
})

# Single element array
errors = [{
  error: "this_is_bad",
  details: "This is not something good?",
}]
```

#### Missing pieces

Method declarations are not included, I played around with adding `Layout/FirstMethodParameterLineBreak`, while it does make it a little better, it does not forces arguments to go in individual lines, unless I'm missing something Rubocop seems to be lacking the required `Layout/MultilineMethodParameterLineBreaks` rule for this.
Because of that, I decided no to touch method declarations at all for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,40 +272,61 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
   ~~~ ruby
   # bad
-  
+
   method(arg_1, arg_2,
     arg_3
   )
-  
+
   [
     value_1, value_2,
     value_3,
   ]
-  
+
   {
     key1: value_1,
     key2: value_2, key3: value_3,
   }
 
   # good
-  
+
   method(
     arg_1,
     arg_2,
     arg_3,
   )
-  
+
   [
     value_1,
     value_2,
     value_3,
   ]
-  
+
   {
     key1: value_1,
     key2: value_2,
     key3: value_3,
   }
+
+  # good (special cases)
+
+  # Single argument method call
+  method({
+    foo: bar,
+  })
+
+  # Last argument, itself is multiline
+  class User
+    after_save :method, if: -> {
+      do_some_checks
+    }
+  end
+
+  # Single value array
+  errors = [{
+    error_code: 1234,
+    error_message: "This is an error",
+  }]
+
   ~~~
 
 * Separate magic comments from code and documentation with a blank line.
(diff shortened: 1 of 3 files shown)


sambostock:
Are there relevant sections in `README.md` that should be updated?

Korri:
> Are there relevant sections in `README.md` that should be updated?

Good point, does that update makes sense? https://github.com/Shopify/ruby-style-guide/pull/387/commits/e0082bb616852a0ff9395c4b51fd6fff2bbd9255

sambostock (approved):
I think validating this by enabling it in some other repos and checking the output makes sense, as we did in Core, which revealed we needed to enable `Layout/FirstArgumentIndentation`.

Assuming we're feeling confident, let's do it 👍

Korri:
Updated this to depend on the changes made [in this PR](https://github.com/rubocop/rubocop/pull/10812), allowing us to keep a few exceptions to the new rules.
This now will have to wait on the next version bump of Rubocop though!

Korri:
@rafaelfranca @sambostock 

Rubocop just release a new version, so this PR is finally ready to go with new exceptions for some single/last elements!

Any objections ?

sambostock (approved):
I've just merged the bump to Rubocop 1.41 (#465), so if you rebase this to get rid of the merge conflict, it should be good to go.

PR #388, opened 2022-05-12
Release 2.6.0

adrianna-chang-shopify:
Release a new version so that consumers get the `IgnoredPatterns` => `AllowedPatterns` config rename.

Draft release here: https://github.com/Shopify/ruby-style-guide/releases/edit/untagged-048380b893a1794a3cfa

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    rubocop-shopify (2.5.0)
+    rubocop-shopify (2.6.0)
       rubocop (~> 1.29)
 
 GEM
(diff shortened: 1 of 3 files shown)


sambostock:
Yeah, as I suspected CI [fails](https://github.com/Shopify/ruby-style-guide/runs/6414507323?check_suite_focus=true) if we run it with the oldest Rubocop version.

sambostock (changes requested):
We should address the Rubocop version issue before releasing the next version. One way to do this would be to merge #389 first, if we're okay with springing a minimum Ruby version on our consumers.

PR #389, opened 2022-05-12
Declare `required_ruby_version` 2.7 and run CI against Ruby 2.7 - 3.1, and minimum Rubocop version

sambostock:
The last time the config was updated (#386), we missed that we were adding a dependency on a more recent version of Rubocop than the one declared as minimum in the `gemspec`, as mentioned in https://github.com/Shopify/ruby-style-guide/pull/388#issuecomment-1125418247.

To address this, and prevent it from happening again, the following changes are made:

- the minimum `rubocop` version is bumped accordingly
- in addition to the version in `Gemfile.lock`, CI now also runs against the version of `rubocop` declared to be the minimum

This more recent version also drops support for Ruby 2.5 and made me realize we should be testing against multiple Ruby versions, so the following other changes are made:

- `required_ruby_version` is declared in `gemspec`
  - `2.7` was chosen as it is the oldest version not at [end-of-life](https://endoflife.date/ruby)
- CI now runs against Rubies `2.7`, `3.0`, and `3.1`
  - a test is added to ensure this list always includes the `required_ruby_version`

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -4,11 +4,12 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: Ruby ${{matrix.gemfile}}
+    name: Ruby ${{matrix.ruby}} / ${{matrix.gemfile}}
 
     strategy:
       matrix:
         gemfile: [Gemfile, gemfiles/minimum_rubocop.gemfile]
+        ruby: ["2.7", "3.0", "3.1"]
 
     env:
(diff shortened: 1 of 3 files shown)


nvasilevski (approved):

rafaelfranca (approved):

volmer:
Sorry I'm late to this but if we always want to use the minimum required version of RuboCop specified on the gemspec for testing then why don't we just declare it directly on the `Gemfile`?

PR #394, opened 2022-05-25
Clarify "Avoid defensive programming"

nunosilva800:
Link to http://www.erlang.se/doc/programming_rules.shtml#HDR11 is broken and this bit might deserve some more explanation.

I think this is a great summary, but here are some other links / resources:
- https://web.archive.org/web/20211013164839/http://www.erlang.se/doc/programming_rules.shtml#HDR11
- https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Code in a functional way. Avoid mutation (side effects) when you can.
 
-* [Avoid defensive programming](http://www.erlang.se/doc/programming_rules.shtml#HDR11).
+* Avoid defensive programming
+  > Overly defensive programming may safeguard against errors that will never be encountered, thus incurring run-time and maintenance costs.
 
 * Avoid mutating arguments.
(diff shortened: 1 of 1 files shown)


tomstuart:
I think the Wikipedia quote is good, but this might be a more controversial change than just fixing the link to point to the archived version of the same page, which says basically the same thing but in more detail. Let’s see what reviewers think!

README.md
```diff
@@ -69,7 +69,8 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 
 * Code in a functional way. Avoid mutation (side effects) when you can.
 
-* [Avoid defensive programming](http://www.erlang.se/doc/programming_rules.shtml#HDR11).
+* Avoid defensive programming
```
nunosilva800, line 72:
```suggestion
* [Avoid defensive programming](https://web.archive.org/web/20211013164839/http://www.erlang.se/doc/programming_rules.shtml#HDR11)
```

PR #422, opened 2022-07-21
Match rubocop-shopify with style guide recommendations

Schwad:
## Background

There is a discrepancy between shopify-rubocop and the Shopify Ruby Styleguide.

The [Ruby Styleguide says:]()

> Use a class << self block over def self. when defining class methods, and group them together within a single block.

( [pr for background](https://github.com/Shopify/ruby-style-guide/pull/29) )

There is an option to enable this in rubocop under `Style/ClassMethodsDefinitions` for `self_class`. Our current default does the opposite, with the `EnforcedStyle` being `def_self`.

This corrects the default.

## For Reviewers

This cop is currently set to `Enabled: false`. If we care enough about this to have an opinion in the styleguide should it be changed to `Enabled: true`? The `def self` vs `class << self` is a long-running question of preference so it might be nice to give clarity on this.

Signed-off-by: Nick Schwaderer <nick.schwaderer@shopify.com>

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -726,4 +726,5 @@ Style/YodaCondition:
 
 Style/ClassMethodsDefinitions:
   EnforcedStyle: self_class
+  Enabled: true
 
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -723,3 +723,7 @@ Style/WordArray:
 
 Style/YodaCondition:
   Enabled: false
+
+Style/ClassMethodsDefinitions:
+  EnforcedStyle: self_class
+
--- a/test/fixtures/full_config.yml
+++ b/test/fixtures/full_config.yml
@@ -2370,7 +2370,7 @@ Style/ClassMethodsDefinitions:
   StyleGuide: "#def-self-class-methods"
   Enabled: false
   VersionAdded: '0.89'
-  EnforcedStyle: def_self
+  EnforcedStyle: self_class
   SupportedStyles:
   - def_self
   - self_class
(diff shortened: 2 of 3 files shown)


sambostock (approved):
This seems fine, given it doesn't actually enable the cop (which would be a bigger discussion, due to the impact), and just configures the cop so that _if enabled_ it would enforce the style consistent with our recommendation.

Nit about the gem name in the description though 😅
```diff
-shopify-rubocop
+rubocop-shopify
```

Schwad:
> Nit about the gem name in the description though 😅

First-time contributor 🙈 I'll get the name right some day! thanks

volmer (commented):
Could you please change the config to `Enabled: true`?

The rule has been added to the Style Guide for years now, so everyone should be writing code that follows it by now, even if RuboCop wasn't able to catch those offences so far. Besides it wouldn't make sense to tweak a cop's config if it's not enabled by default.

joeldrapper:
Is there a way to enforce this style only when there's more than one class method?

csalvato:
> Is there a way to enforce this style only when there's more than one class method?

For the sake of consistency, I don't think that makes sense, personally.

I actually prefer `def self.` in general, but if we are going to have an opinion on the style, it should be consistent even with just one class method, imo.

promulo (approved):

andyw8:
Should we set this to ignore migrations?

https://buildkite.com/shopify/shopify-graphiql-app/builds/2155#0183c3f3-710f-424a-9fbb-7c6f30ca2979

rafaelfranca:
That isn't true since Rails 3.1. I don't think we should care about very old Rails versions.

PR #432, opened 2022-08-19
Enable TrailingCommaInArguments

nunosilva800:
Sets the same behaviour for trailing comma on all scenarios:
- arguments
- array literals
- hash literals

This particular cop was disabled in https://github.com/Shopify/ruby-style-guide/pull/18
with the main argument being the option `EnforcedStyleForMultiline` was binary (no comma vs comma).

Having the `consistent_comma` possibility, I think we can please everyone (🤞).

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -698,7 +698,7 @@ Style/TrailingBodyOnMethodDefinition:
   Enabled: false
 
 Style/TrailingCommaInArguments:
-  Enabled: false
+  EnforcedStyleForMultiline: consistent_comma
 
 Style/TrailingCommaInArrayLiteral:
   EnforcedStyleForMultiline: consistent_comma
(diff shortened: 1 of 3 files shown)


nunosilva800:
Why is this a good thing? Having a trailing comma on the last argument of a multi-line argument list means that the next time an argument is added to that list, results in a 1-line change, instead of a 2 line change. Imo this makes code reviewing a tiny bit simpler.

rafaelfranca:
What is the impact of this rule in Shopify/shopify?

sambostock:
> I'll fix the CI offenses if this gets approval. 

For what it's worth, there is only a single offense in this repo and it's auto-correctable, so I'd just get fixing it out of the way, that way this PR is ready to go if we decide to accept it.

nunosilva800:
> What is the impact of this rule in Shopify/shopify?

I'm surprised this didn't work: https://github.com/Shopify/shopify/compare/trailing-comma-args

I can't get this cop to fail in anyway 🤔

rafaelfranca:
You also need `Enabled: true`

nunosilva800:
Thanks! Here are the numbers: 

> 49115 files inspected, 143321 offenses detected, 143321 offenses autocorrectable

PR #444, opened 2022-09-13
Set Style/TrailingCommaInArguments's EnforcedStyleForMultiline to comma

rafaelfranca:
Differently from the Hash and Array versions of this rule, this one uses the existence of parentheses to determine if a comma is needed or not, not if the method is multiline. As our style guide require parentheses for all method calls, so this rule can't use the consistent_comma option.


rafaelfranca:
We had it disabled. I think comma is the one that makes most sense if we are to enable it.

PR #456, opened 2022-11-11
Enforce single-quotes instead of double-quotes.

burke:
I'm bracing for a storm here.

I think our style guide has `Style/StringLiterals` exactly backwards. I think enforcing double quotes for uninterpolated literals is fairly aesthetically offensive to almost anyone that's been writing ruby for more than a handful of years. This has bothered me since it was introduced in 2021 but after ranting about this with @tobi and @duncan yesterday I figured I'd toss up this PR.

The obvious downside of this PR as is, is that it effectively forces a rewrite of probably near a million lines of code for aesthetic reasons—but hey, what is code if not purely aesthetics, right?

I could be argued down to not having an enforced opinion on this at all, and I suspect that's actually the right call given the inertia we have here right now.


rafaelfranca:
There is a point to be made that double quotes everywhere is the style of the [Rails project for years now](https://github.com/rails/rails/blob/46608c9398a71b68e37ded3d7875bdc9c214f876/.rubocop.yml#L187) and all code generated by Rails will have that style out-of-box. As many other things at Shopify, we try to follow the Rails standards.

> I could be argued down to not having an enforced opinion on this at all, and I suspect that's actually the right call given the inertia we have here right now.

That is definitely an option here specially since that doesn't require us to rewrite millions of lines of code. If we want to do something, I think that would be the best course of action.

But again, this proposal doesn't bring anything that wasn't already considered in past discussions. I'm not sure we should be doing anything.

burke:
> There is a point to be made that double quotes everywhere is the style of the [Rails project for years now](https://github.com/rails/rails/blob/46608c9398a71b68e37ded3d7875bdc9c214f876/.rubocop.yml#L187) and all code generated by Rails will have that style out-of-box. As many other things at Shopify, we try to follow the Rails standards.

Aaaahhhh this is new information to me. Yeah, clearly we can't change our stance to enforce single-quotes.

I do kind of still think this should be unenforced in that case 🤔

rafaelfranca:
I'd be fine with not enforcing this rule, but there is a large camp of people that just don't want to make a choice and follow a style-guide. That is what the style-guide here to provide. That is the direction that Go, Rust, Elixir went. There is a value of a decision being made for people that don't really feel strongly if they should enforce double quotes or not, they just want to write code.

I do think as a company that employs a team responsible for setting the direction of where Ruby and Rails should go, inside and outside the company, should make a choice on this matter, and trust that team know what is best for Ruby and Rails. I don't want every team when starting a project to decide if they want to enforce this or not, or realizing after a while that nothing is enforced and asking themselves why isn't enforced.

But, [as I said in the other discussion](https://github.com/Shopify/ruby-style-guide/pull/238#issuecomment-786113234), enforcing this role in the style-guide allow us to document a way forward, but it is ok for teams to make a call if they want to enforce this rule or not in their projects if that is what they want to do.

tszymczyszyn-shopify:
> I think enforcing double quotes for uninterpolated literals is fairly aesthetically offensive to almost anyone that's been writing ruby for more than a handful of years

Any data to support this opinion? 

My own opinion is that the use of single quotes is counterproductive, as adding/removing interpolation requites changing quote type.

PR #474, opened 2023-01-02
Disable Style/OperatorMethodCall and release 2.11.1

Korri:
This rule is causing (at least one) syntax error in code by replacing:
```ruby
T.unsafe(self).==(...)
```
by
```ruby
T.unsafe(self) == (...)
```

Which is not valid ruby

### DIFF
--- a/rubocop-shopify.gemspec
+++ b/rubocop-shopify.gemspec
@@ -3,7 +3,7 @@
 Gem::Specification.new do |s|
   s.platform    = Gem::Platform::RUBY
   s.name        = "rubocop-shopify"
-  s.version     = "2.11.0"
+  s.version     = "2.11.1"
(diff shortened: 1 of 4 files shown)


rubocop-shopify.gemspec
```diff
@@ -3,7 +3,11 @@
 Gem::Specification.new do |s|
   s.platform    = Gem::Platform::RUBY
   s.name        = "rubocop-shopify"
+<<<<<<< HEAD
```
sambostock, line 6: Merge conflict markers

sambostock (approved):
Looks good, other than the rogue merge conflict markers. We can merge after that's fixed.

sambostock (approved):

sambostock:
Fixed in https://github.com/rubocop/rubocop/pull/11378

PR #496, opened 2023-02-09
Allow `def test_...` lines to be long

sambostock:
We currently allow lines like

```ruby
test "a really really ... really long test name" do
```

using the Active Support declarative test style to violate `Layout/LineLength`. It stands to reason we should also allow the equivalent Minitest style

```ruby
def test_a_really_really_..._really_long_test_name
```


nvasilevski (approved):
👍

nunosilva800:
Good stuff!

rafaelfranca (approved):

PR #498, opened 2023-02-14
Explore plugin configs

sambostock:
This PR explores the idea of us also providing configs for other RuboCop plugins. Currently teams at Shopify configure these independently, but maybe we want to take a stance on some of the cops.

> **Warning**
> Work in progress.

- RuboCop config requires only the RuboCop plugins whose gemspec is loaded
- Each plugin has a corresponding file of defaults which are inherited
- Labels are attached automatically to PRs touching each config file

### TBD:
- Should we have a minimal config? Should we keep the existing config as is, and instead add an "everything.yml" config?
- Should we test compatibility with _just_ RuboCop and no plugins?
- What cops to enable? Maybe disallowing "pending" shouldn't apply to plugins, so we don't break consumers?


rubocop.minitest.yml
```diff
@@ -0,0 +1,2 @@
+Minitest:
+  Enabled: true
```
sambostock, line 2: I haven't bothered filling out most of these files, but presumably they'd look like:

```yaml
PluginName:
  Enabled: true

# ...config for cops where we want to diverge from the default...
```

rubocop.yml
```diff
@@ -3,6 +3,22 @@ inherit_mode:
...
+    "rubocop-performance" => "rubocop.performance.yml",
+    "rubocop-rails" => "rubocop.rails.yml",
+    "rubocop-rake" => "rubocop.graphql.yml",
+    "rubocop-rspec" => "rubocop.rspec.yml",
+  }.select { |plugin_name, _| Gem.loaded_specs.include?(plugin_name) }
+%>
+
+require: <%= plugin_configs.keys.to_json %>
+
+inherit_from: <%= plugin_configs.values.to_json %>
```
sambostock, line 20: This is where the magic happens. 🪄 ✨ 👮‍♂️

Are there reasons why this would be a bad idea?

### Pros

- Allows us to inject config without forcing the dependency on any plugins, or knowing anything fancy about the host app.
- Removes the need to separate gems (e.g. `rubocop-rails-shopify` or whatever).
- Removes the need for consumers to opt-into the configs themselves. They can simply follow the suggestions from Rubocop, and as soon as they add a plugin it will "magically" be configured.

### Cons

- Doesn't allow us to enforce the version of the plugins.
  - This might get tricky as cops are added/removed.
    - We would probably have to fallback to enforcing the version in ERB, which isn't ideal.
- Could there be cases where it's unsafe to enable these?
  - We could move the magic into a special file (e.g. `infer-rubocop.yml`), and keep this file simple, but that would mean consumers would have to know to switch.
  - Maybe it's sufficient to require consumers to simply override the config if it isn't compatible.

sambostock: I've tentatively moved this into `rubocop.infer.yml`, since this repo itself ran into inapplicable cops. Although that may be a special case, so we can always put it back. If we went with this separate file, then it would mean consumers could opt into the magic

```yaml
inherit_gem:
  rubocop-shopify: rubocop.infer.yml
```

or stick with the current minimal approach:

```yaml
inherit_gem
  rubocop-shopify: rubocop.yml
```

or take a custom approach

```yaml
inherit_gem
  rubocop-shopify:
    - rubocop.yml
    - rubocop.rails.yml
```

rubocop.minitest.yml
sambostock: We could alternatively call these after the gem names (e.g. `rubocop-minitest.yml`). I just went with what we use in Core.

rubocop.infer.yml
```diff
@@ -0,0 +1,16 @@
...
+  # Identify installed RuboCop plugins to configure
+  plugin_configs = {
+    "rubocop" => "rubocop.yml",
+    "rubocop-graphql" => "rubocop.graphql.yml",
+    "rubocop-minitest" => "rubocop.minitest.yml",
+    "rubocop-performance" => "rubocop.performance.yml",
+    "rubocop-rails" => "rubocop.rails.yml",
+    "rubocop-rake" => "rubocop.graphql.yml",
+    "rubocop-rspec" => "rubocop.rspec.yml",
+  }.select { |plugin_name, _| Gem.loaded_specs.include?(plugin_name) }
```
rafaelfranca, line 11: The tricky part is that this checks if the gem is installed in the computer, not in the bundler right?

rubocop.minitest.yml
rafaelfranca: I prefer to not use defaults. Let's always curate cops.

rubocop.infer.yml
sambostock: Yes. The way it would work is that if a consumer has

```yml
# .rubocop.yml
inherit_gem:
  rubocop-shopify: rubocop.infer.yml
```

then when they run `rubocop` it will look at their installed gems, and inherit from the applicable configs.

rafaelfranca: But if the app doesn't have rubocop-minitest for example but the person has it installed globally we should not use it.

We should be using what is on the Gemfile.lock of the application, not what is on the gem paths.

sambostock: Yes, good point. I should have said "in their Gemfile" not "installed". I just confirmed `Gem.loaded_specs` handles this the way we want:

```console
% cat Gemfile
source 'https://rubygems.org'
% bundle exec ruby -e 'puts Gem.loaded_specs.key?("rubocop-minitest")'
false
% gem install rubocop-minitest
Successfully installed rubocop-minitest-0.32.2
Parsing documentation for rubocop-minitest-0.32.2
Installing ri documentation for rubocop-minitest-0.32.2
Done installing documentation for rubocop-minitest after 0 seconds
1 gem installed
% bundle exec ruby -e 'puts Gem.loaded_specs.key?("rubocop-minitest")'
false
% bundle add rubocop-minitest > /dev/null && cat Gemfile
source 'https://rubygems.org'

gem "rubocop-minitest", "~> 0.32.2"
% bundle exec ruby -e 'puts Gem.loaded_specs.key?("rubocop-minitest")'
true
% bundle exec ruby -e 'puts Gem.loaded_specs.key?("rubocop")'
true
```

PR #523, opened 2023-04-02
Skip dumping config if required token is missing

sambostock:
We need a specific token to be able to make commits that will trigger other actions. However, Dependabot commits use different secrets and our token automation does not (yet) configure those.

Therefore, this makes dumping the config conditional on having the token. It is structured as a separate job due to limitations around the availability of secrets in `if` contexts, and to avoid having the conditional on every step.

### DIFF
--- a/.github/workflows/dump-config.yml
+++ b/.github/workflows/dump-config.yml
@@ -3,8 +3,22 @@ name: Dump config
 on: [push, pull_request]
 
 jobs:
+  check-token:
+    name: Check for token
+    runs-on: ubuntu-latest
+    outputs:
+      token-held: ${{ steps.set_output.outputs.token_held }}
+    steps:
+      - name: check for token
+        id: set_output
+        run: '[ -z "$TOKEN_HELD" ] || echo "token_held=true" >> "$GITHUB_OUTPUT"'
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dump-config.yml
```diff
@@ -3,8 +3,22 @@ name: Dump config
...
+    name: Check for token
+    runs-on: ubuntu-latest
+    outputs:
+      token-held: ${{ steps.set_output.outputs.token_held }}
+    steps:
+      - name: check for token
+        id: set_output
+        run: '[ -z "$TOKEN_HELD" ] || echo "token_held=true" >> "$GITHUB_OUTPUT"'
+        env:
+          TOKEN_HELD: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
```
sambostock, line 16: I tried a secret that doesn't exist, and [successfully skipped dumping the config](https://github.com/Shopify/ruby-style-guide/actions/runs/4589514822).

```yaml
TOKEN_HELD: ${{ secrets.EXAMPLE }}
```

sambostock:
@robthornton Valid point, but I don't know what the timeline is on the token tooling change, and until then I figure the automation still has value for non-Dependabot commits. This change allows those to work, and is forward compatible if we do add a Dependabot secret token.

PR #529, opened 2023-04-13
Enable Style/ReturnNil cop

larouxn:
Proposing we enable `Style/ReturnNil` to enforce `return if/unless condition` style over `return nil unless condition` as the `nil` component is redundant/unnecessary. Let's make this consistent.

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ReturnNil

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -756,6 +756,9 @@ Style/RegexpLiteral:
 Style/RescueStandardError:
   Enabled: false
 
+Style/ReturnNil:
+  Enabled: true
+
 Style/SelectByRegexp:
   Enabled: false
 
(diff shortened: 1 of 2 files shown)


sambostock (commented):
I'm not necessarily opposed to this, but some people might prefer the explicitness. 🤷 

Can you run Rubocop against Core with this setting and see how many offenses there are?

rafaelfranca (approved):
I'm being more conservative on enabling cops that can have impact on existing code and doesn't add much value other than consistency, but I'm ok with this cop being enabled.

larouxn:
👋 @sambostock, I was finally able to run this cop against core. `1594` infractions. Results [here](https://github.com/Shopify/shopify/pull/420116).

sambostock (approved):

PR #530, opened 2023-04-14
Try dumping config on `pull_request_target` instead

sambostock:
This may have access to the necessary secrets.


sambostock:
https://github.com/Shopify/ruby-style-guide/actions/runs/4702515585/jobs/8339874458?pr=531 seems promising, but we'll have to wait until the next Rubocop release that touches the default config to be sure, since that would actually exercise trying to generate a commit.

PR #539, opened 2023-05-10
Memoization strategy

jmeinerz:
Memoization is useful if you want to run a command once but be able to access that value many times.

However, the popular `||=` operator is not great for when your command might return a falsey value. By guarding memoization instance methods with `if defined?`, we make sure it only runs once.

This is particularly important to stop expensive queries that may return `nil` or `false` from running needlessly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1407,6 +1407,6 @@ end
 def merchant
   return @merchant if defined?(@merchant)
   
-  @merchant ||= Merchant.find_by(id: merchant_id)
+  @merchant = Merchant.find_by(id: merchant_id)
 end
 ~~~
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * [Regular Expressions](#regular-expressions)
 * [Percent Literals](#percent-literals)
 * [Testing](#testing)
+* [Memoization](#memoization)
 
 ## General
 
@@ -1391,3 +1392,21 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
     do_something
   end
   ~~~
+  
+## Memoization
+
+* Prefer `return @x if defined?(@x)` over a simple `||=`
+
+~~~ ruby
+# bad - if the method is called N times, the query will also run N times
+def merchant
+  @merchant ||= Merchant.find_by(id: merchant_id)
+end
+
+# good - A single query, even when @merchant is nil
+def merchant
+  return @merchant if defined?(@merchant)
+  
+  @merchant ||= Merchant.find_by(id: merchant_id)
+end
+~~~
--- a/README.md
+++ b/README.md
@@ -1398,7 +1398,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Prefer `return @x if defined?(@x)` over a simple `||=`
 
 ~~~ ruby
-# bad - if the method is called N times, the query will also run N times
+# bad - if @merchant is assigned a falsy value (`nil`, `false`, `0`, etc) and is called N times, the query will also run N times
 def merchant
   @merchant ||= Merchant.find_by(id: merchant_id)
 end
--- a/README.md
+++ b/README.md
@@ -1398,7 +1398,7 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
 * Prefer `return @x if defined?(@x)` over a simple `||=`
 
 ~~~ ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1391,3 +1392,21 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
...
+# bad - if the method is called N times, the query will also run N times
+def merchant
+  @merchant ||= Merchant.find_by(id: merchant_id)
+end
+
+# good - A single query, even when @merchant is nil
+def merchant
+  return @merchant if defined?(@merchant)
+  
+  @merchant ||= Merchant.find_by(id: merchant_id)
```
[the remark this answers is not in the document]
jmeinerz: Agreed :+1:

README.md
```diff
@@ -1391,3 +1392,21 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
     do_something
   end
   ~~~
+  
+## Memoization
+
+* Prefer `return @x if defined?(@x)` over a simple `||=`
+
+~~~ ruby
+# bad - if @merchant is assigned a falsy value (`nil`, `false`, `0`, etc) and is called N times, the query will also run N times
```
tjwp, line 1401: Only `nil` and `false` are falsy in Ruby:

```suggestion
# bad - if @merchant is assigned a falsy value (`nil` or `false`) and is called N times, the query will also run N times
```

README.md
```diff
@@ -1391,3 +1392,21 @@ documentation](https://docs.rubocop.org/rubocop/configuration.html#inheriting-co
     do_something
   end
   ~~~
+  
+## Memoization
+
+* Prefer `return @x if defined?(@x)` over a simple `||=`
```
cbothner, line 1398:
```suggestion
* Prefer `return @x if defined?(@x)` over `||=` if the statement being memoized can evaluate to `nil` or `false`.
```

cbothner (commented):
Can we add a little nuance to the recommendation? Most of the time the terser rose memoization is appropriate.

README.md
cbothner: Frankly, everything around memoization requires nuance, because memoization is caching and cache invalidation is one of the Two Hard Problems[^1]. I think your addition is a good detail to include in our style guide to reduce the chance someone mindlessly repeats the rose memoization pattern. But it would also be suboptimal for someone to mindlessly repeat `return @x if defined?(@x)`.

[^1]: along with naming things and off-by-one errors, famously

I don't support _forbidding_ rose memoization, so I'll advocate for the nuance in our discussion of it.

---

On a similar note, a section on Memoization might also include a note about memoizing a method that takes arguments.

```
# Bad. A stale value can be returned when a different argument is passed for subsequent calls
def expensive_result(input)
  @expensive_result ||= ExpensiveResult.calculate(input)
end

# Better. Recalculates when a different argument is passed.
# Consider using an LRU or not memoizing at all if the argument cardinality
# is high or the object is long-lived.
def expensive_result(input)
  @expensive_results ||= {}
  @expensive_results[input] ||= ExpensiveResult.calculate(input)
end
```

rafaelfranca (changes requested):
There is overhead of using `defined?`. Since it is a keyword, it is hard to optimize by the VM. In reality, memoization itself is hard to optimize in the VM right now.

If I'd write anything in the styleguide about memoization would be in the lines of telling people to avoid memoization. Judging by the code I see in our applications, most of the time memoization is done at the wrong time and in the wrong place.

But regardless, `defined?` over `||=` isn't a matter of style, they have very different behavior and both can and should be used, so I don't think a style guide is the right place to include an explanation of when use one over the other. This is more content for a Ruby book, or a Ruby tutorial document. People need to learn when to use one over the other, or when not use any of them. This isn't the job of this styleguide.

jmeinerz:
> There is overhead of using defined?. Since it is a keyword, it is hard to optimize by the VM. In reality, memoization itself is hard to optimize in the VM right now.

If both are hard to optimise, is there really a drawback in suggesting one over the other?

> If I'd write anything in the styleguide about memoization would be in the lines of telling people to avoid memoization. Judging by the code I see in our applications, most of the time memoization is done at the wrong time and in the wrong place.

This is an interesting angle. I'm not opposed to it at all, but I think there's still value in nudging people to a way with fewer side effects.

> People need to learn when to use one over the other, or when not use any of them. This isn't the job of this styleguide.

In my view, it does fit the styleguide. The way I think about it is that there's more harm to come from leaving this completely open than to make a loose suggestion for `defined?`. And I do agree with you that "people need to learn" but I also think we should try and support those who are earlier in their journey to write good code without having to know *everything*. For a junior developer, a day at Shopify is already a lot to take, and I would like to help them write code with fewer side effects.

In your opinion, what is the goal of the styleguide?

PR #199, opened 2020-09-28
Mark Style/FrozenStringLiteralComment as safe to autocorrect

KaanOzkan:
There was internal discussion regarding FrozenStringLiteralComment cop being deemed unsafe to autocorrect in Rubocop. This recent change threw people off that expect `-a` to fix FrozenStringLiteralComment errors as it has done in the past, eg. https://github.com/Shopify/rubocop-sorbet/issues/46. 

This past behaviour along with the assumption that we expect developers to write code that is frozen string safe lead to the decision of keeping the existing practice of `-a` autocorrecting this cop, even though it was marked as unsafe to autocorrect in rubocop.

Rubocop PRs that made autocorrect unsafe: https://github.com/rubocop-hq/rubocop/pull/7307 & https://github.com/rubocop-hq/rubocop/pull/8529

PR #589, opened 2023-11-09
Enforce `frozen_string_literal: true`

sambostock:
The default `EnforcedStyle` is `always`, which enforces that all files have a `# frozen_string_literal` comment, but doesn't care if it's `true` or `false`. `always_true` enforces `true`, which we want. Any existing files can simply be added to a TODO list or disable the cop.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -548,6 +548,7 @@ Style/FormatStringToken:
 
 Style/FrozenStringLiteralComment:
   SafeAutoCorrect: true
+  EnforcedStyle: always_true
   Details: 'Add `# frozen_string_literal: true` to the top of the file. Frozen string
     literals will become the default in a future Ruby version, and we want to make
     sure we''re ready.'


rafaelfranca (approved):

sambostock:
Hmmm... wouldn't that be fine, since the developer would have to triage it in their PR anyway? I think if we have that false, then the dev would have to manually add the comment, rather than the something like VS Code adding it automatically.

sambostock:
Arguably, it was always possible to write that file without a comment, and run the autocorrect which would default to `frozen_string_literal: true` I believe, which would be the same problem. The only difference being that modifying it on existing files could break production code, but it would go through PR review, so should be caught.

andyw8:
Ok, yeah, it's probably not a big deal, just feels a little off to label it as safe when it's not.

sambostock:
Yeah, I agree. I tracked down the [PR](https://github.com/Shopify/ruby-style-guide/pull/199) that originally made that change though, and it looks like it's intentional.

PR #607, opened 2024-01-25
Allow long lines on spec titles too

sambostock:
We allow long test titles of the form

```ruby
def test_really_long
```
```ruby
test "really long" do
```

This extends the `Regexp` to support `Minitest::Spec` and `RSpec` style tests

```ruby
describe "really long" do
  context "really long" do
    it "really long" do
```

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -73,7 +73,7 @@ Layout/LineEndStringConcatenationIndentation:
 
 Layout/LineLength:
   AllowedPatterns:
-  - "\\A\\s*(remote_)?test(_\\w+)?\\s.*(do|->)(\\s|\\Z)"
+  - "\\A\\s*((remote_)?test|context|describe|it|pending|skip|xit)(_\\w+)?\\s.*(do|->)(\\s|\\Z)"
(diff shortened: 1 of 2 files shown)


rubocop.yml
```diff
@@ -73,7 +73,7 @@ Layout/LineEndStringConcatenationIndentation:
 
 Layout/LineLength:
   AllowedPatterns:
-  - "\\A\\s*(remote_)?test(_\\w+)?\\s.*(do|->)(\\s|\\Z)"
+  - "\\A\\s*((remote_)?test|context|describe|it|pending|skip|xit)(_\\w+)?\\s.*(do|->)(\\s|\\Z)"
```
sambostock, line 76: As a counter-point, maybe we _shouldn't_ do this because the "spec" style of tests would prefer nesting contexts to having a long description... 🤔

```ruby
context "really" do
  context "really" do
    it "long" do
```

rafaelfranca:
RSpec and Minitest::Spec aren't things we should use at Shopify. The styleguide should not encourage them by allowing them to be used better. If anything we should be actively working to remove their usage

PR #612, opened 2024-02-12
Reduce minimum RuboCop version back to `~> 1.51`

sambostock:
Since we're introducing the

```yaml
<% if rubocop_version >= "X.YZ" %>
Dept/Cop:
  Enabled: true
<% end %>
```

approach to configuring RuboCop without requiring a specific version of the gem (ahead of supporting plugin configs), we should just use that approach to avoid changing the minimum required version in the next release.

### ⚠️ Before Merging

- [x] Rebase after #613 is merged.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     rubocop-shopify (2.14.0)
-      rubocop (~> 1.53)
+      rubocop (~> 1.51)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


sambostock:
CI is failing with `comparison of Gem::Version with String failed`, but that [should work](https://github.com/rubygems/rubygems/blob/5cedfc8ad923cbc679049ed69fbb1baf1753617f/lib/rubygems/version.rb#L362)... 🤔

sambostock:
It turns out `Gem::Version#<=>` only _relatively_ recently learned to accept `String` arguments, but we can backport the guardclause to be able to use it.

rafaelfranca (approved):

PR #621, opened 2024-03-06
Use Prism parser engine

sambostock:
Once Prism support is stable in RuboCop, we'll likely want to default to enabling it.

As of https://github.com/rubocop/rubocop/releases/tag/v1.62.0, support is experimental, so we don't want to merge this yet.

### DIFF
--- a/rubocop-shopify.gemspec
+++ b/rubocop-shopify.gemspec
@@ -23,5 +23,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.7.0"
 
-  s.add_dependency("rubocop", "~> 1.51")
+  s.add_dependency("prism")
+  s.add_dependency("rubocop", "~> 1.62")
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,8 @@ PATH
   remote: .
   specs:
     rubocop-shopify (2.15.1)
-      rubocop (~> 1.51)
+      prism
+      rubocop (~> 1.62)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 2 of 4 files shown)


rubocop-shopify.gemspec
```diff
@@ -23,5 +23,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.7.0"
 
-  s.add_dependency("rubocop", "~> 1.51")
+  s.add_dependency("prism")
```
paracycle, line 26: I think in the final form, we would want a lower bound on the version of Prism, to make sure that we have all the bugfixes related to RuboCop translation layer in.

sambostock: Yeah, agreed. I was a little surprised [`rubocop-ast` doesn't require a minimum version](https://github.com/rubocop/rubocop-ast/pull/285#discussion_r1509359810). It seems like something you'd want to do if you _know_ it won't work with versions below some minimum.

paracycle: Yeah, agreed. I think once we are past v1.0 of Prism, then we should be able to add that as a minimum version. otherwise, my concern is that folks who try RuboCop with Prism as a backend on vanilla Ruby 3.3 will have a bad time, since the version of Prism released with Ruby 3.3 will be horribly out of date.

sambostock: Yeah, agreed. RuboCop's bug report template asks the submitted to include the output of `rubocop -V` which would include the Prism version, but I don't understand why not simply restrict the version and prevent bug reports against outdated Prism versions from being opened in the first place.

PR #674, opened 2025-01-22
Disable Style/MethodCallWithArgsParentheses cop for migrations

obie:
Same rationale as disabling this cop for Gemfile (the source is a DSL and reads worse with parentheses)

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -635,6 +635,7 @@ Style/MethodCallWithArgsParentheses:
   - puts
   Exclude:
   - "/**/Gemfile"
+  - "/**/db/migrate/*"
 
 Style/MinMax:
   Enabled: false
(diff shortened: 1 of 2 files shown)


rubocop.yml
```diff
@@ -635,6 +635,7 @@ Style/MethodCallWithArgsParentheses:
   - puts
   Exclude:
   - "/**/Gemfile"
+  - "/**/db/migrate/*"
```
paracycle, line 638: Hmm, I am not sure if a blanket exclusion for migration files is the best thing to do. We want to exclude only because things like `create_table`, etc read better as DSLs, but people are not limited to writing those in migration files, and we probably **do** want to lint all other method calls.

sambostock:
Using an example migration from the Rails docs:

```ruby
# db/migrate/20240502100843_create_products.rb
class CreateProducts < ActiveRecord::Migration[8.0]
  def change
    create_table :products do |t|
      t.string :name
      t.text :description

      t.timestamps
    end
  end
end
```

I believe "macros" are loosely defined as methods called within a `module`/`class` body:

```ruby
class Post < ApplicationRecord
  belongs_to :user # macro
end

User = Class.new(ApplicationRecord)
User.has_many(:posts) # not macro
```

By that definition, nothing in that migration is a "macro".

Issue #675, opened 2025-01-23
Enable `Lint/DuplicateCaseCondition`?


andyw8:
Noticed in https://github.com/Shopify/spoom/pull/632

It's enabled by default in RuboCop but disabled in our config.

There are 5 offences in Core.

rafaelfranca:
Yes. We can denable this one.

PR #677, opened 2025-01-23
Enable `Lint/DuplicateCaseCondition`

andyw8:
Closes https://github.com/Shopify/ruby-style-guide/issues/675


andyw8:
Should we make a new release? It's almost a year since the last one.

PR #687, opened 2025-02-13
The great reset

rafaelfranca:
A change of direction for this project.

Before, we would restrict the expressiveness of Ruby for the sake of consistency and safety. This was causing Ruby programs that felt not really like Ruby.

Example of those rules were:

* mandating parentheses in all method call
* forcing some methods over others
* trailing commas.

Now, this style guide, at least its cops, will focus on providing some consistency without making Ruby not look like Ruby. It will be more open to interpretation, and provide the freedom that Ruby always wanted to provide to their users.

It will also focus on catching mistakes. But not ones that require us to change how we write Ruby program.

In this PR:

* We are disabling all cops with the exception of:
  - Style/FrozenStringLiteralComment - At least until those are made default in Ruby, we will still want this comment
  - Some cops in the Bundler and Gemspec department - Those catch errors that doesn't change how we write Ruby

* Drop support to Ruby 3.0

Later PRs will enable cops in other department, after careful review of their usefulness.


burke (approved):
I think this is right. We will definitely find we want to add some of them back but let's feel that pain first.

PR #693, opened 2025-02-19
Allow RBS comments without a leading space

Morriar:
Follow up on https://github.com/Shopify/rubocop-sorbet/pull/282#issuecomment-2669493082.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -110,6 +110,11 @@ Layout/HashAlignment:
 Layout/HeredocIndentation:
   Enabled: false
 
+<% if rubocop_version >= "1.67" %>
+Layout/LeadingCommentSpace:
+  AllowRBSInlineAnnotation: true
+<% end %>
+
 Layout/LineContinuationLeadingSpace:
   Enabled: true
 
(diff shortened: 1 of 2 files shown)


rubocop.yml
```diff
@@ -110,6 +110,10 @@ Layout/HashAlignment:
 Layout/HeredocIndentation:
   Enabled: false
 
+Layout/LeadingCommentSpace:
+  Enabled: true
```
rafaelfranca, line 114:
```suggestion
```

rubocop.yml
```diff
@@ -110,6 +110,9 @@ Layout/HashAlignment:
 Layout/HeredocIndentation:
   Enabled: false
 
+Layout/LeadingCommentSpace:
+  AllowRBSInlineAnnotation: true
```
sambostock, line 114: To enable this config without bumping the minimum RuboCop version, we can use the same technique as we did to configure new cops:

```suggestion
<% if rubocop_version >= "1.67" %>
Layout/LeadingCommentSpace:
  AllowRBSInlineAnnotation: true
<% end %>
```

sambostock: On the other hand, _The Great Reset_ certainly warrants a major version bump, so we could take the opportunity to bump the minimum RuboCop version.

Thoughts @rafaelfranca?

rafaelfranca: We can do that later.

PR #699, opened 2025-02-28
Add Lint/NoReturnInMemoization cop

rafaelfranca:
Cop based on https://docs.rubocop.org/rubocop/cops_lint.html#lintnoreturninbeginendblocks, but with the difference that it only cares about instance variables, given sometimes using return inside `begin..end` blocks is idiomatic Ruby, like when handling exceptions.

Example of such case:

```ruby
def some_method
  some_value = begin
    get_external_api
  rescue ApiError
    return { error: "Api returned an error" }
  end

  do_something(some_value)
end
```

This cop checks for the use of `return` with a value in `begin..end` blocks in the context of instance variable assignment such as memoization. Using `return` with a value in these blocks can lead to unexpected behavior as the `return` will exit the method and not set the value of the instance variable.

```ruby
# bad
def foo
  @foo ||= begin
    return 1 if bar
    2
  end
end

# bad
def foo
  @foo = begin
    return 1 if bar
    2
  end
end

# bad
def foo
  @foo += begin
    return 1 if bar
    2
  end
end

# bad - using return in rescue blocks still exits the method
def foo
  @foo ||= begin
    bar
  rescue
    return 2 if baz
  end
end

# good
def foo
  @foo ||= begin
    if bar
      1
    else
      2
    end
  end
end

# good - not an assignment to an instance variable
def foo
  foo = begin
    return 1 if bar
    2
  end
end

# good - proper exception handling without return
def foo
  @foo ||= begin
    bar
  rescue
    baz ? 2 : 3
  end
end
```

### DIFF
--- a/lib/rubocop/cop/lint/no_return_in_memoization.rb
+++ b/lib/rubocop/cop/lint/no_return_in_memoization.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Lint
+      # Checks for the use of a `return` with a value in `begin..end` blocks
+      # in the context of instance variable assignment such as memoization.
+      # Using `return` with a value in these blocks can lead to unexpected behavior
+      # as the `return` will exit the method that contains and not set the values
+      # of the instance variable.
+      #
+      # @example
+      #
+      #   # bad
+      #   def foo
+      #     @foo ||= begin
+      #       return 1 if bar
+      #       2
+      #     end
+      #   end
+      #
+      #   # bad
+      #   def foo
+      #     @foo = begin
+      #       return 1 if bar
+      #       2
+      #     end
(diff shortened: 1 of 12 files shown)


lib/rubocop/cop/lint/no_return_in_memoization.rb
```diff
@@ -0,0 +1,97 @@
...
+          end
+        end
+
+        def on_op_asgn(node)
+          return unless node.assignment_node.ivasgn_type?
+
+          on_or_asgn(node)
+        end
+
+        alias on_ivasgn on_or_asgn
```
amomchilov, line 93: `&&=` is niche, but makes sense to cover here, too.

```suggestion
        alias on_ivasgn on_or_asgn
        alias on_and_asgn on_or_asgn
```

amomchilov (commented):
Nice.

Issue #737, opened 2025-06-04
RuboCop Warning Due to Deprecated Cop Name in rubocop-shopify


gbs4ever:
We're encountering the following warning during RuboCop execution:

Warning: Using `Naming/PredicateName` configuration in /.rubocop.yml for `Naming/PredicatePrefix`.

This is due to a recent change in RuboCop where the cop Naming/PredicateName was renamed to Naming/PredicatePrefix. The warning is being triggered by the rubocop-shopify gem, which still references the deprecated cop name in its internal .rubocop.yml configuration.

sambostock:
https://github.com/Shopify/ruby-style-guide/pull/736/commits/ee416ca43e62cd9ad5d53124e09d008506bfad4a in #736 will handle this by conditionally configuring the correctly named cop. We may want to cherry-pick that commit to backport the fix to the `2-stable` branch.

gbs4ever:
Thank you for the fix @sambostock

PR #739, opened 2025-06-05
Prepare for 2.17.1 release

sambostock:
This includes #738, which addresses #737.


rubocop-shopify.gemspec
```diff
@@ -3,7 +3,7 @@
 Gem::Specification.new do |s|
   s.platform    = Gem::Platform::RUBY
   s.name        = "rubocop-shopify"
-  s.version     = "2.17.0"
+  s.version     = "2.17.1"
```
sambostock, line 6: I figured a patch version bump is adequate since we're just addressing a cop rename warning.

PR #770, opened 2025-12-01
Enable `Lint/EmptyBlock`

cocoahero:
Hopefully a straight-forward proposal to enable `Lint/EmptyBlock`, in an effort to prevent situations like this:

```ruby
field :my_graphql_field, :string, null: true do |f|
  # something was here but removed by rudimentary automation
end
```

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -229,7 +229,7 @@ Lint/DuplicateSetElement:
   Enabled: true
 
 Lint/EmptyBlock:
-  Enabled: false
+  Enabled: true
 
 Lint/EmptyClass:
   Enabled: false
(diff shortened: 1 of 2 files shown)


sambostock (approved):
Ironically, your example in the PR description _would_ be allowed, because it contains a comment (but I _think_ that is what we want, for cases where for some reason passing an empty block is required).

derik-thiessen (approved):

rafaelfranca (changes requested):
As I suspected, there are method that are today expected to use empty blocks, so we can't enable this rule.

One example is the `on_mutation` in graphql-easy-testing. You not always want to provide a block, but the library requires a block to be passed. So while we don't fix those requirement for empty blocks, being they legit or not, we can't enable this rule.

PR #797, opened 2026-07-15
Add cop to disallow case structures where all when clauses are proc literals

rwstauner:
An if/else tree is equivalent without allocating the procs.

This also triggers if the only non-proc clauses are literal values.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    rubocop-shopify (3.0.1)
+    rubocop-shopify (3.0.2)
       lint_roller
       rubocop (~> 1.72, >= 1.72.1)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -2,3 +2,8 @@ Lint/NoReturnInMemoization:
   Enabled: true
   VersionAdded: "3.0.0"
   Description: "Checks for the use of a `return` with a value in `begin..end` blocks in the context of instance variable assignment such as memoization."
+
+Style/ProcCaseWhen:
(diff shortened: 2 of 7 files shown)


sambostock:
This _feels_ like one of those things where maybe we should be taking a look at solving it at the interpreter level. Is there a reason an `if`/`elsif`/`else` chain can avoid the allocation, but a `case` can't?

lib/rubocop/cop/lint/proc_case_when.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Lint
```
rafaelfranca: I'd say this is more the `Style` department than `Lint`. Yes, it is about the allocation and the `call` overhead, but in term of style, the `if/else` are more natural.

rwstauner: Done

rwstauner:
> This feels like one of those things where maybe we should be taking a look at solving it at the interpreter level. Is there a reason an if/elsif/else chain can avoid the allocation, but a case can't?

The only reason a proc is used here is to bend the condition to fit into the case statement.  Since you wouldn't do that using if/else, there's no special handling needed when using the more natural syntax.

It would be _possible_ to optimize this in the interpreter (I believe we can reduce the allocation count by caching the proc at the parser level), but in order to inline the block to remove all the overhead I'm not convinced the interpeter bloat needed to accomplish the task is worth it.

rafaelfranca (approved):

PR #801, opened 2026-08-07
Enable Style/RedundantFreeze cop

rafaelfranca:
This cop checks usages of `Object#freeze` on immutable objects.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -1196,9 +1196,6 @@ Style/RedundantFilterChain:
 Style/RedundantFormat:
   Enabled: false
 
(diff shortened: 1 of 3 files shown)


sambostock (approved):

rubocop.yml
```diff
@@ -1197,7 +1197,7 @@ Style/RedundantFormat:
   Enabled: false
 
 Style/RedundantFreeze:
-  Enabled: false
+  Enabled: true
```
sambostock:
```suggestion
```

I think this will address the redundant config CI check

PR #809, opened 2026-09-14
Disable Style/ReduceToHash

hmcguire-shopify:
```
ary.to_h { [k, v] }
```

allocates N arrays vs.

```
ary.reduce({}) { _1[_2] = v }
```

does not.


nvasilevski:
Does this mean there is an opportunity for an optimization? Like a special instruction in VM? Because at least on a napkin I want to claim that if behavior is the same at least in theory we should be able to achive the same performance while keeping the prettier option

nvasilevski (approved):
I have nothing against disabling it except that I'd like our long-term desire to be to have `.to_h` form to avoid allocation or potentially have a different allocation-less API to turn enumerable into a hash. It's not like `reduce` or `each_with_object` options are ugly but just less readable

sambostock (approved):

rafaelfranca:
It is possible to optimize in the VM, but it is too hard to do properly https://github.com/ruby/ruby/compare/master...Shopify:ruby:to_h-allocation-zjit

rafaelfranca (approved):

